PR #244, opened 2021-06-09
New Storybook Github Workflow for Oazo AWS account

eskp:


.github/workflows/storybook.yml
```diff
@@ -8,9 +8,9 @@ jobs:
     name: Deploy storybook
     runs-on: ubuntu-latest
     env:
-      REGION: eu-central-1
-      BUCKET_NAME: storybook.oasis.app
-      CLOUDFRONT_ID: E1PL3BMP98DRBZ
+      REGION: us-east-1
+      BUCKET_NAME: storybook.oazolabs.xyz
```
jeannettemcd, line 12: This bucket name is using the "test" DNS in it's name, when we switch to `oasis.app` this bucket will be destroyed and a new one created...  Is there contents that needs to be copied over ? Any thing we need to be aware ?

eskp: Good point, afaik we'd just need to redeploy from this repo and the bucket will get repopulated but could confirm with the devs.

dizzy (approved):

PR #270, opened 2021-06-22
Rz/infra migration banner

pacxiu:
# [Infra migration banner](https://app.clubhouse.io/oazo-apps/story/2095/infra-transition-communication-about-transition-to-users)
  
## Changes 👷‍♀️
- adding infra migration banner + "Visit the old Oasis"
- extract `WelcomeAnnouncement` (might be removed or have copy changed after we remove infra banner)
  
image

  
## How to test 🧪
- run UI, check if banner is visible for all pages
    
## Definition of done ✔️
- [ ] ~~Unit tests written where needed and passing~~
- [ ] ~~End-to-end tests for happy path~~
- [ ] ~~Documentation updated <When applicable>~~

### DIFF
--- a/components/Announcement.tsx
+++ b/components/Announcement.tsx
@@ -1,7 +1,9 @@
 import { Icon } from '@makerdao/dai-ui-icons'
 import { WithChildren } from 'helpers/types'
 import React from 'react'
-import { Flex, SxProps } from 'theme-ui'
+import { Flex, Grid, SxProps, Text } from 'theme-ui'
+
+import { AppLink } from './Links'
 
 export function Announcement({ children, sx }: WithChildren & SxProps) {
   return (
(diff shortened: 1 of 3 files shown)


peculiarity:
1. Agree with @ferni  regarding the margin under mobile
2. I thought that we should leave the previous message as well
3. The link in oasis-app to the old borrow is purple and in here it's not
4. Not quite sure if this should be visible on the wallet connect page. Please double check with Chris.

wallet connect

pacxiu:
1. Fixed in both repos
2. For now we should have just 1 banner and have `Visit the old Oasis` inside this one  - so we don't have 2 banners (feedback from Discord). After removing this infra banner old one will be restored, which is hidden now, maybe with copy changed.
3. Unified should be text color in both
4. Added boxShadow here, should be everywhere

PR #282, opened 2021-06-28
Dp/swap price

paszkowskiDamian:
# [Calculations 1st iteration](https://app.clubhouse.io/oazo-apps/story/2081/add-support-for-the-buying-power-and-price-impact-calculations-needed-in-the-onboarding)

## Changes 👷‍♀️
- fetching price data from 1inch
- 1st iteration of calculations 
  
## How to test 🧪
- go to open multiply vault
- play with values
    
## Definition of done ✔️


features/openMultiplyVault/openMultiplyVaultCalculations.ts
```diff
@@ -43,61 +45,174 @@ export const defaultOpenVaultStateCalculations: OpenMultiplyVaultCalculations =
...
-  const afterNetValueUSD = totalExposureUSD?.minus(afterOutstandingDebt) || zero
   const afterNetValue = afterNetValueUSD.div(currentCollateralPrice)
 
   const afterLiquidationPrice = afterCollateralizationRatio.gt(0)
     ? currentCollateralPrice.times(liquidationRatio).div(afterCollateralizationRatio)
     : zero
 
+  const impact =
+    quote?.status === 'SUCCESS'
+      ? new BigNumber(quote.daiAmount.minus(quote.collateralAmount)).div(quote.daiAmount)
```
pacxiu, line 204: Shouldn't that be `times` instead of `minus`?

features/openMultiplyVault/components/OpenMultiplyVaultEditing.tsx
```diff
@@ -118,10 +117,12 @@ export function OpenMultiplyVaultEditing(props: OpenMultiplyVaultState) {
               background: sliderBackground,
             }}
             disabled={!canAdjustRisk}
-            step={2}
-            value={multiply?.toNumber() || 0}
+            step={5}
+            min={0}
+            max={100}
```
pacxiu, line 122: I'm not sure if maybe we wouldn't want to have `max` set to like `1000` so we could better control over steps amount so it will be a little bit smoother.

And then in calculations it would be `slider.div(100)` - so maybe extracting this `100` as easily editable const value? Just was thinking as Henry wanted to have 50 steps so it will be smooth for users.

paszkowskiDamian: I can decrease step if you want. This is also a part where I actually wondering how to handle it. Right now some generic value 0 - 100 is mapped to range maxPossibleCollRatio and liquidationRatio. We could actually set slider values to min=liqRatio and max=maxPossible, but I wasn't sure of the impact on the slider that it would have.

pacxiu: We can let Henry and Hazel play with it and eventually tune it I think

pacxiu (approved):

pacxiu (approved):

PR #318, opened 2021-07-19
Add responsive styles for Header, VaultPage and VaultForm

pacxiu:
# [Vault Components RWD](https://clubhouseLink)

partial design link - https://www.figma.com/file/EdkPKn73UBzSbkr1xKwo6c/App?node-id=2032%3A21287

## Changes 👷‍♀️
- change breakpoints from `640`, `832`, `1024` to `768`, `896`, `1024` which is more reasonable for real devices
- rename `GeneralManageVault` to uppercase
- bring back storybook
- fix overflow on modals (body was cut off)
- fix additional space taken by `svg` with token icons definitions
- remove `100vw` on html, which was causing slight difference `8px` off on some of the containers, add `ModalHTMLOverflow` component which handles preventing jumping on Modal opening
- add mobile navigation for connected state:
image
- unify `AppSpinner` + animation for Open, OpenMultiply, Manage Views
- add `hover` effect on desktop for triggering tooltips, and `onClick` for mobile (have to be tested on actual mobile device, eg. by visiting deployment url)
- add responsivness to VaultHeader and layout for Open, OpenMultiply, Manage Views
image
- introduce `useBreakpointIndex` for conditional rendering components if needed, forked from [ThemeUI](https://github.com/system-ui/theme-ui/blob/develop/packages/match-media/src/index.ts) (out current theme ui version is not letting us use this package directly)
- introduce `SharedUIProvider` for toggling sidebar
- add modal like sidebar with banner for Open, OpenMultiply, Manage Views + slide in animation from right
image

## How to test 🧪
- investigate UI on mobile `375-1024px` for vault (have to consult with designers how we should approach rwd in general for app, seems like it wasn't that much thought of)
    
## Definition of done ✔️

###
For Modal like Sidebar with vault I've decided to introduce this `SharedUIProvider`. Problem is that with our current `useModal` we have to explicitly pass <ModalLikeComponent> to onClick handler with it's props. That would cause our header to have logic based on urls to set proper text and proper modal content + props for those vault sidebars.

With this provider we are setting proper text + eventually just trigger open/close and render proper sidebar if needed in the actual page/component, which I find more natural in this case.


paszkowskiDamian:
Jul-27-2021 11-25-30

components/vault/VaultHeader.tsx
```diff
@@ -32,25 +38,29 @@ function VaultIlkDetailsItem({
...
+            isTouchDevice
+              ? () => {
+                  setTooltipOpen(!tooltipOpen)
+                }
+              : undefined
+          }
+          sx={{ fontSize: '0px', cursor: 'pointer' }}
+        >
+          <Icon name="tooltip" color="primary" size="auto" width="14px" height="14px" />
+        </Box>
```
paszkowskiDamian, line 61: How about changing target to entire Box (icon + label)?  It would be easer on mobile devices to open modal.

paszkowskiDamian (dismissed):

PR #401, opened 2021-09-08
Add feedback form.

ferni:
# [Add feedback form (Typeform) script to homepage](https://app.clubhouse.io/oazo-apps/story/2593/add-feedback-form-typeform-script-to-homepage)

## Changes 👷‍♀️

- Added Typeform (a feedback form widget) to the homepage.
  
It must be tested using https , or it doesn't work. Test deployment: https://oasis-borrow-ft-feedbac-hi2vev.herokuapp.com/

### DIFF
--- a/features/landing/LandingView.tsx
+++ b/features/landing/LandingView.tsx
@@ -15,11 +15,11 @@ import { staticFilesRuntimeUrl } from 'helpers/staticPaths'
 import { Trans, useTranslation } from 'next-i18next'
 import getConfig from 'next/config'
 import React, { ComponentProps, useCallback } from 'react'
-import { useEffect } from 'react'
 import { Box, Button, Card, Flex, Grid, Heading, Image, SxStyleProp, Text } from 'theme-ui'
 import { fadeInAnimation, slideInAnimation } from 'theme/animations'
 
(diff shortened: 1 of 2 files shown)


features/landing/LandingView.tsx
```diff
@@ -214,6 +215,29 @@ function LandingCards() {
   )
 }
 
+function TypeformWidget() {
```
peculiarity, line 218: Can you extract this as a separate component?

peculiarity (approved):

features/landing/TypeformWidget.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+  useEffect(() => {
+    if (document.querySelector(`script#${scriptId}`) !== null) {
+      return
+    }
+    const script = document.createElement('script')
+    script.id = scriptId
+    script.src = '//embed.typeform.com/next/embed.js'
+    document.body.append(script)
+  })
```
pacxiu, line 14: One question, why isn't it mounted only once, with `useEffect` and `[]` is there some kind of race condition?

I see that there is `return` but still, we invoke `document.querySelector....` on every render

ferni: Ahh, right, will add the dependencies array, thanks!

pacxiu (approved):

PR #413, opened 2021-09-10
Add fee calcs for closeTo actions

pacxiu:
## Changes 👷‍♀️
- add fee calcs for close to flows
  
## How to test 🧪
- inspect UI/Storybook
    
## Definition of done ✔️

### DIFF
--- a/helpers/multiply/calculations.ts
+++ b/helpers/multiply/calculations.ts
@@ -209,11 +209,14 @@ export type CloseToParams = {
   fromTokenAmount: BigNumber
   toTokenAmount: BigNumber
   minToTokenAmount: BigNumber
+  oazoFee: BigNumber
+  loanFee: BigNumber
 }
 
 export function calculateCloseToDaiParams(
   marketPrice: BigNumber,
   OF: BigNumber,
+  FF: BigNumber,
   currentCollateral: BigNumber,
   slippage: BigNumber,
 ): CloseToParams {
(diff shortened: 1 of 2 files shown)


helpers/multiply/calculations.ts
```diff
@@ -250,5 +256,7 @@ export function calculateCloseToCollateralParams(
     fromTokenAmount,
     toTokenAmount,
     minToTokenAmount,
+    oazoFee: currentDebt.times(OF),
```
adamskrodzki, line 259: this one for sure:
oazoFee:expectedFinalDebt.times(OF)

this I'm not 100% sure of 
loanFee:expectedFinalDebt.times(FF) @fenris85 ?

adamskrodzki (changes requested):
while difference is very small and most likely we can get away with it

I found out some inconsistencies in:
 helpers/multiply/calculations.ts 
 
 Also I advice that for FlashLoan fee @fenris85  also will take a look

helpers/multiply/calculations.ts
pacxiu: `expectedFinalDebt` already has `OF` and `FF` inside so we would be applying `OF` on top of those.

If `loanFee` here is wrong it should also be wrong for `closeToDai` as they are currently calculated the same:
```
loanFee: currentDebt.times(one.plus(OF)).times(FF)
```

adamskrodzki: close to collateral is little different than close to DAI if it comes to OAZO fee since it is taken at different moment (after a swap not before) and from different amount (all existing collateral not debt)

if it comes to OF situation in close to collateral it is as follows. 

in order to have enough DAI at the end of operation
- currentDebt*(1+FF)  required to repay FlashLoan

exchaange must have had after 1inch operation, following amount 

currentDebt*(1+FF) /(1-OF)   

and this is amount from which OazoFee was taken



(debtDelta+flashLoanFee)  and this is value after oazo fee is applied

adamskrodzki (approved):

peculiarity (approved):

PR #443, opened 2021-09-14
Make a simple announcement that will be visible on all app pages and …

peculiarity:
According to a requirement we should have a very easy to configure announcement that anyone can active.
- The announcement **MUST** contain a text.
- The announcement **MUST** contain link to discord.  
  - This link should be configurable because the invitation links expire so the person who 
    configures the announcement should be able easily provide a new link.
- The announcement **COULD** have another **OPTIONAL** link. 
- The announcement **MUST** be closeable per session. 
  - If the user during his session closed the announcement on any page, they shouldn't see it on any other pages during that session. 
  - If the user has announcement dismissed and closes the tab where the application is loading and then opens the application in a new tab, the announcement MUST be visible, even thought they have closed it in a previous session.
 - The announcement **MUST** be visible only on the Landing page  + All application related pages.
 - The announcement **MUST NOT** be visible on Connect Page, Team Page, Terms page, Support page and these sort of pages.
 - The announcement **MUST** look nice on mobile devices and everything should be rendered without any overlapping elements.

### DIFF
--- a/components/Layouts.tsx
+++ b/components/Layouts.tsx
@@ -9,11 +9,17 @@ import { Background } from 'theme/Background'
 import { BackgroundLight } from 'theme/BackgroundLight'
 import { BackgroundLighter } from 'theme/BackgroundLighter'
 
+import { GenericAnnouncement } from './Announcement'
+
 interface BasicLayoutProps extends WithChildren {
   header: JSX.Element
   footer?: JSX.Element
   sx?: SxStyleProp
   variant?: string
+}
+
+interface WithAnnouncementLayoutProps extends BasicLayoutProps {
+  showAnnouncement: boolean
 }
 
 export function BasicLayout({ header, footer, children, sx, variant }: BasicLayoutProps) {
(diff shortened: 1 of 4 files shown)


components/Layouts.tsx
```diff
@@ -35,16 +41,58 @@ export function BasicLayout({ header, footer, children, sx, variant }: BasicLayo
...
+        ...sx,
+      }}
+    >
+      {header}
+      {showAnnouncement && (
+        <Container variant="announcement">
+          <GenericAnnouncement
+            text="Welcome to the new Oasis.app. We are thrilled to have you here. Please check the new stuff. How long can it go is that I
+        m curious"
+            discordLink="https://discord.com/channels/837076147694207067/846839026651889675"
```
adamskrodzki, line 67: https://discord.gg/Kc2bBB59GC

adamskrodzki (approved):
It works ok for me,
I updated discord link

image

two questions (but i think it is not for now)

x closing mark is off if banner is not wide

this color is bearly visible

PR #462, opened 2021-09-16
Redeclare icons since they got lost after a merge

peculiarity:
There are some missing icons on products.


adamskrodzki (approved):
Looks good, 

icon is indeed missing on prod and present on feature branch, 
I do not think this change can have any more impact

PR #476, opened 2021-09-21
Add Goerli Support

peculiarity:
Add support for Goerli network


blockchain/config.ts
```diff
@@ -53,9 +54,6 @@ const protoMain = {
     ...getCollateralTokens(mainnetAddresses),
     WETH: contractDesc(eth, mainnetAddresses['ETH']),
     DAI: contractDesc(erc20, mainnetAddresses['MCD_DAI']),
-    MKR: contractDesc(erc20, mainnetAddresses.MCD_GOV),
```
adamskrodzki, line 56: why some mainnet related stuff gets removed here ?

peculiarity: These are leftovers from previous projects. We don't need MKR in oasis-borrow. Ive decided that since I'm doing changes in this file, I might clean some stuff as well otherwise it might be confusing for other people

johnnieskywalker (approved):

adamskrodzki (approved):

PR #493, opened 2021-09-28
As/2734 base fee based calculation

adamskrodzki:
Change pipelines in a way described in 
https://app.shortcut.com/oazo-apps/story/2734/change-gasprice-estimator-to-use-basefee-and-set-max-priority-fee-per-gas-to-5-gwei


blockchain/prices.ts
```diff
@@ -18,16 +18,41 @@ export interface Ticker {
   [label: string]: BigNumber
 }
 
-export type GasPrice$ = Observable<BigNumber>
+export type GasPriceParams = {
+  MaxFeePerGas: BigNumber
```
pacxiu, line 22: Too keep naming consistent on object params across app it should start from lowercase so:
```
maxFeePerGas: BigNumber
maxPriorityFeePerGas: BigNumber
```

also small typo: `Pririty`

blockchain/prices.ts
```diff
@@ -18,16 +18,41 @@ export interface Ticker {
...
+      const retVal = {
+        MaxFeePerGas: new BigNumber(block.baseFeePerGas).multipliedBy(2).plus(minersTip),
+        MaxPrirityFeePerGas: minersTip,
+      } as GasPriceParams
+      console.log(
+        `GasPriceParams = { MaxFeePerGas:${retVal.MaxFeePerGas}, MaxPrirityFeePerGas:${retVal.MaxPrirityFeePerGas} }`,
+      )
+      return retVal
+    }),
+    distinctUntilChanged(
```
pacxiu, line 52: I think you can use `isEqual` from lodash, so it will be just `distinctUntilChanged(isEqual)`

pacxiu:
I think there will also need to be some changes inside `doGasEstimation` function to calculate estimated fees as it also uses `gasPrice$`

I guess TS haven't caught that as it fails on Prettier from what I see

blockchain/prices.ts
adamskrodzki: Actually there are, in 
helpers/form.ts

pacxiu (approved):

PR #495, opened 2021-09-29
Add slippage limit settings

pacxiu:

## Changes 👷‍♀️
- item 1
  
## How to test 🧪
- step 1 ...
    
## Definition of done ✔️

### DIFF
--- a/features/manageMultiplyVault/manageMultiplyVaultEnvironment.ts
+++ b/features/manageMultiplyVault/manageMultiplyVaultEnvironment.ts
@@ -1,6 +1,7 @@
 import { IlkDataChange } from 'blockchain/ilks'
 import { VaultChange } from 'blockchain/vaults'
 import { PriceInfoChange } from 'features/shared/priceInfo'
+import { SlippageChange } from 'features/slippageLimit/slippageLimit'
 
 import { BalanceInfoChange } from '../shared/balanceInfo'
 import { ManageMultiplyVaultChange, ManageMultiplyVaultState } from './manageMultiplyVault'
@@ -10,6 +11,7 @@ export type ManageVaultEnvironmentChange =
   | BalanceInfoChange
   | IlkDataChange
   | VaultChange
+  | SlippageChange
 
 export function applyManageVaultEnvironment(
   change: ManageMultiplyVaultChange,
@@ -40,6 +42,13 @@ export function applyManageVaultEnvironment(
     return {
       ...state,
       vault: change.vault,
+    }
+  }
+
+  if (change.kind === 'slippage') {
+    return {
+      ...state,
+      slippage: change.slippage.div(100),
     }
   }
 
--- a/features/openMultiplyVault/openMultiplyVault.ts
+++ b/features/openMultiplyVault/openMultiplyVault.ts
@@ -7,8 +7,8 @@ import { ExchangeAction, Quote } from 'features/exchange/exchange'
 import { calculateInitialTotalSteps } from 'features/openVault/openVaultConditions'
 import { BalanceInfo, balanceInfoChange$ } from 'features/shared/balanceInfo'
 import { PriceInfo, priceInfoChange$ } from 'features/shared/priceInfo'
+import { slippageChange$, SlippageLimitState } from 'features/slippageLimit/slippageLimit'
 import { GasEstimationStatus, HasGasEstimation } from 'helpers/form'
-import { SLIPPAGE } from 'helpers/multiply/calculations'
 import { curry } from 'lodash'
 import { combineLatest, iif, merge, Observable, of, Subject, throwError } from 'rxjs'
 import { first, map, scan, shareReplay, switchMap, tap } from 'rxjs/operators'
@@ -295,6 +295,7 @@ export function createOpenMultiplyVault$(
     action: ExchangeAction,
   ) => Observable<Quote>,
   addGasEstimation$: AddGasEstimationFunction,
+  slippageLimit$: Observable<SlippageLimitState>,
   ilk: string,
 ): Observable<OpenMultiplyVaultState> {
   return ilks$.pipe(
@@ -302,9 +303,9 @@ export function createOpenMultiplyVault$(
       iif(
         () => !ilks.some((i) => i === ilk),
         throwError(new Error(`Ilk ${ilk} does not exist`)),
-        combineLatest(context$, txHelpers$, ilkData$(ilk)).pipe(
+        combineLatest(context$, txHelpers$, ilkData$(ilk), slippageLimit$).pipe(
           first(),
-          switchMap(([context, txHelpers, ilkData]) => {
+          switchMap(([context, txHelpers, ilkData, { slippage }]) => {
             const { token } = ilkData
             const account = context.account
             return combineLatest(
@@ -349,7 +350,7 @@ export function createOpenMultiplyVault$(
                       errorMessages: [],
                       warningMessages: [],
                       summary: defaultOpenVaultSummary,
-                      slippage: SLIPPAGE,
+                      slippage: slippage.div(100),
                       totalSteps,
                       currentStep: 1,
                       exchangeError: false,
@@ -364,8 +365,9 @@ export function createOpenMultiplyVault$(
                       priceInfoChange$(priceInfo$, token),
                       balanceInfoChange$(balanceInfo$, token, account),
                       createIlkDataChange$(ilkData$, ilk),
-                      createInitialQuoteChange(exchangeQuote$, token),
+                      createInitialQuoteChange(exchangeQuote$, token, slippage.div(100)),
                       createExchangeChange$(exchangeQuote$, stateSubject$),
+                      slippageChange$(slippageLimit$),
                     )
 
                     const connectedProxyAddress$ = proxyAddress$(account)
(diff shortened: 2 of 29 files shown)


scripts/docker-compose.yml
```diff
@@ -27,27 +27,27 @@ services:
...
-      - DATABASE_URL=${DATABASE_URL}
-    ports:
-      - '3000:3000'
+  # borrow:
+  #   image: oasisdexorg/oasis-borrow:latest
+  #   container_name: oasis-borrow
+  #   depends_on:
+  #     - postgres
+  #     - localnode
+  #   command:
```
peculiarity, line 36: This shouldn't be commented

features/slippageLimit/slippageLimitLocal.ts
```diff
@@ -0,0 +1,14 @@
+import BigNumber from 'bignumber.js'
+import { Observable, of } from 'rxjs'
+
+const SLIPPAGE_LOCAL_STORAGE = 'slippage_setting'
+
+export function checkSlippageLocalStorage$(): Observable<{ slippage: string | null }> {
```
peculiarity, line 6: Don't get implementational details in the names. Today is localStorage, tomorrow might be something else and you will have to rename the method. 
`getSlippage` and `storeSlippage` is enough

features/slippageLimit/slippageLimitLocal.ts
```diff
@@ -0,0 +1,14 @@
...
+  const slippage = localStorage.getItem(SLIPPAGE_LOCAL_STORAGE)
+  return of({ slippage })
+}
+
+export function saveSlippageLocalStorage$(slippage: BigNumber): Observable<void> {
```
peculiarity, line 11: The returned value is undefined and not void :D

features/slippageLimit/slippageLimit.ts
```diff
@@ -0,0 +1,161 @@
...
+      }
+
+      const slippage = initialSlippage ? new BigNumber(initialSlippage) : SLIPPAGE_DEFAULT
+
+      const initialState: SlippageLimitState = {
+        stage: 'editing',
+        slippage,
+        slippageInput: slippage,
+        setSlippageLow: () => change({ kind: 'slippageInput', slippageInput: SLIPPAGE_LOW }),
+        setSlippageMedium: () => change({ kind: 'slippageInput', slippageInput: SLIPPAGE_MEDIUM }),
```
peculiarity, line 126: I don't think you need a separate method for each case. If we decide to remove one or add another one, we have to change that logic in the pipelines and change types and tests and what not. `setSlippage: (slippage: BigNumber) => change({ kind: 'slippageInput', value: slippage}`. 
You will export these `SLIPPAGE_LOW, SLIPPAGE_MEDIUM, SLIPPAGE_HIGH` and use the in the UI. Instead of 4 functions, you will have 1 and if you decide to extend it, you can only add `SLIPPAGE_SOME_OTHER_VALUE` and use it in the UI.

features/slippageLimit/slippageLimitLocal.ts
peculiarity: Also if you want to return something, return boolean not void

features/slippageLimit/slippageLimit.ts
```diff
@@ -0,0 +1,161 @@
...
+  const invalidSlippage =
+    !slippageInput ||
+    slippageInput.isNaN() ||
+    (slippageInput && (slippageInput.gt(new BigNumber(20)) || slippageInput.lt(new BigNumber(0.1))))
+
+  if (invalidSlippage) {
+    errors.push('invalidSlippage')
+  }
+
+  if (!invalidSlippage && slippageInput.gt(new BigNumber(5))) {
```
peculiarity, line 101: Extract this hardcoded value somewhere. `MAX_SLIPPAGE_LIMIT` or smth like that. You can put it with the otheres `SLIPPAGE_LOW` ..

features/slippageLimit/slippageLimit.ts
```diff
@@ -0,0 +1,161 @@
...
+
+function validate(state: SlippageLimitState): SlippageLimitState {
+  const { slippageInput } = state
+  const errors: SlippageLimitErrorMessages[] = []
+  const warnings: SlippageLimitWarningMessages[] = []
+
+  const invalidSlippage =
+    !slippageInput ||
+    slippageInput.isNaN() ||
+    (slippageInput && (slippageInput.gt(new BigNumber(20)) || slippageInput.lt(new BigNumber(0.1))))
```
peculiarity, line 95: Extract the hardcoded value `MAX_SLIPPAGE_LIMIT` or smth and `MIN_SLIPPAGE_LIMIT`

features/slippageLimit/SlippageLimitView.tsx
```diff
@@ -0,0 +1,224 @@
...
+          {t('slippage-limit.preset-title')}
+        </Text>
+        <Text variant="paragraph3" sx={{ color: 'text.subtitle' }}>
+          {t('slippage-limit.preset-description')}
+        </Text>
+        <AppLink href="https://kb.oasis.app/help/slippage" sx={{ fontSize: 2, fontWeight: 'body' }}>
+          {t('slippage-limit.read-more')}
+        </AppLink>
+        <Flex sx={{ mt: 3 }}>
+          <SlippageOptionButton option={SLIPPAGE_OPTIONS[0]} onClick={setSlippageLow} />
```
peculiarity, line 139:
```js
[SLIPPAGE_LOW, SLIPPAGE_MEDIUM, SLIPPAGE_HIGH].map((value, index) => {
   <SlippageOptionButton key={index} option={SLIPPAGE_OPTIONS[0]} onClick={() => setSlippage(value)} />
})
```

peculiarity: IF you want to extend it , just add another value. If you want to contract it ...remove a value and that will be the only change

features/slippageLimit/SlippageLimitView.tsx
```diff
@@ -0,0 +1,224 @@
...
+    case 'invalidSlippage':
+      return 'slippage-limit.errors.invalid-slippage'
+    case 'highSlippage':
+      return 'slippage-limit.warnings.high-slippage'
+    default:
+      throw new UnreachableCaseError(message)
+  }
+}
+
+function SlippageLimitErrors({
```
peculiarity, line 70: So it's errors but it also accepts warnings? That's little bit smelly. Why do't you create  2 separate components?

features/openMultiplyVault/openMultiplyVaultEnvironment.ts
```diff
@@ -31,5 +36,12 @@ export function applyOpenVaultEnvironment(
     }
   }
 
+  if (change.kind === 'slippage') {
+    return {
+      ...state,
+      slippage: change.slippage.div(100),
```
peculiarity, line 42: I think we have `formatPercent` for that..

features/openMultiplyVault/openMultiplyVault.ts
```diff
@@ -349,7 +350,7 @@ export function createOpenMultiplyVault$(
                       errorMessages: [],
                       warningMessages: [],
                       summary: defaultOpenVaultSummary,
-                      slippage: SLIPPAGE,
+                      slippage: slippage.div(100),
```
peculiarity, line 353: extract this into a block variable. YOu are doing the same division twice.

features/manageMultiplyVault/manageMultiplyVaultEnvironment.ts
```diff
@@ -43,5 +45,12 @@ export function applyManageVaultEnvironment(
     }
   }
 
+  if (change.kind === 'slippage') {
+    return {
+      ...state,
+      slippage: change.slippage.div(100),
```
peculiarity, line 51: If you are dividing this everywhere why don't you returend a number which is actually a percent ?

peculiarity (changes requested):
Please check my comments

features/slippageLimit/SlippageLimitView.tsx
pacxiu: I've just changed name to `Messages` they are not that different so it doesn't make sense to duplicate those.

features/slippageLimit/slippageLimitLocal.ts
pacxiu: I'll just add that as a parameter to pipeline itself. Implementation details is okay, as eg. for termsAcceptance/vault we have `local` and `Api` and we just pass it directly as pipeline dependency, this way we can switch between those.

features/slippageLimit/slippageLimit.ts
pacxiu: We did it same way for allowances (deposit, unlimited, custom .etc), maybe with some slight difference on `change.kind` and it was done exactly for that case so we could write test for all preset options that we have.

Then indeed when there is modification on those methods, some test will fail - but that would be expected and tbh nice.

peculiarity: I really don't like this approach. It's not flexible and  you should test behavior, you shouldn't test state and what I mean by that. There is not different if you will set 0.5, 1, 2. What it matters is that the value you pass is set. This what the test shoud look like. 
What is the difference between `setSlippageLow` , `setSlippageMedium` and `setSlippageHigh`? It's only the value you pass. I can pass the same values from the custom input... There is no thing special in these 3 methods.

adamskrodzki: I think biznes-wise allowance is a different case zero, deposit, unlimited, custom  are logically different, and there will be no new logical state of allowance, You cannot say the same about Slippage

Honestly speaking I do not understand, why this enumeration is even needed here it should get just any number, and somwhere else there should be some enumeration applied in front-end as a property or something

PR #550, opened 2021-10-28
Lw/story/2893/save network

johnnieskywalker:

## Changes 👷‍♀️
- item 1
  
## How to test 🧪
- step 1 ...
    
## Definition of done ✔️

### DIFF
--- a/server/database/migrations/003-chain-id.sql
+++ b/server/database/migrations/003-chain-id.sql
@@ -1,0 +1,2 @@
+ALTER TABLE vault
+ADD COLUMN chain_id INTEGER;
--- a/blockchain/vaults.ts
+++ b/blockchain/vaults.ts
@@ -61,7 +61,9 @@ export function createVaults$(
 
       return fetchVaultIds().pipe(
         switchMap((ids) =>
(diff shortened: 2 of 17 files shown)


server/database/migrations/003-chain-id.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE vault
+ADD COLUMN chain_id INTEGER;
```
adamskrodzki, line 2: what with existing entries ?

johnnieskywalker: There's no simple way to determine which network are on. So the network will be added once the existing entry will be modified. If that's not enough I could probably create some script that would search all existing vaults in blockchains and add network. For ones that exist on couple of networks ... the probably the safest way would be to save them and manually determine for these edge cases. Not sure if this would bring much-added value in contrary to the amount of time for preparing such a script.

The way it works now there's a risk that the vault that id exists on one network will be modified on another and the record in db will be changed. However, when one on another network will be modified new entry with the correct network will be created. Probably the safest way would be to have separate instances of db for each network. This could be also overengineering. Hard to determine how would complexity and amount of edge cases grow with time...

PR #582, opened 2021-11-15
Cookie Banner

ferni:
# [Create Cookie banner](https://app.shortcut.com/oazo-apps/story/2994/create-cookie-banner)

## Changes 👷‍♀️

- Add cookie banner to ~landing page~ all pages.
- Can enable/disable Mixpanel (analytics cookie). Marketing cookie is yet to be implemented (adroll).
- Extract ChevronUpDown component, for all your chevron up/down needs.
- Improve how buttons are shown in Storybook.


johnnieskywalker:
when rejected 
image

johnnieskywalker:
When one goes to app to other page then landing, using for example URL somebody may copy paste it from somewhere or have a bookmark in browser then cookieBanner doesn't show up on other pages and cookieSetting in localStorage is not existing

features/landing/LandingView.tsx
```diff
@@ -321,6 +322,7 @@ export function LandingView() {
       </WithErrorHandler>
       <LandingCards />
       <TypeformWidget />
+      <CookieBanner />
```
johnnieskywalker, line 325: This CookieBanner should appear in all pages that are reachable by URL. User may have for ex bookmarked URL to vault or send some URL to someone else etc. CookieBanner should appear there too. 
So AFAIK besides LandingPage it should be added to all pages in /`oasis-borrow/pages` directory and maybe in couple other Vault views.

features/landing/CookieBanner.tsx
```diff
@@ -0,0 +1,184 @@
...
+    setSettingsAreSaved(true)
+  }
+
+  function rejectCookies() {
+    COOKIE_NAMES.forEach((cookieName) => manageCookie[cookieName].disable())
+    saveSettings({ accepted: false, cookieSettings: initCookieSettings(false) })
+  }
+
+  function acceptCookies() {
+    // enable/disable based on settings
```
johnnieskywalker, line 101: [minor] So maybe instead comment other name of function would be better idea? Maybe `saveAcceptedCookies` or `saveAcceptedAndRejectedCookies`, `saveCookiesSettings` ? Something like this

johnnieskywalker: Ah I got this name because it's related to acceptButton. Maybe it should be just `acceptSelectedCookies` ? 
Well one may unselected both and click accept so actually nothing will be accepted but I guess for method name it will be fine.
image

features/landing/CookieBanner.tsx
```diff
@@ -0,0 +1,184 @@
...
+        gridTemplateColumns: '1fr 1fr',
+        gap: 2,
+        alignItems: 'start',
+        direction: ['rtl', 'unset'],
+      }}
+    >
+      <Button variant="bean" sx={{ fontSize: 2 }} onClick={() => rejectCookies()}>
+        {t('landing.cookie-banner.reject')}
+      </Button>
+      <Button variant="beanActive" sx={{ fontSize: 2 }} onClick={() => acceptCookies()}>
```
johnnieskywalker, line 126: [minor] Method name suggestion `acceptSelectedCookies()`

johnnieskywalker (changes requested):
Everything works as expected and implementation is correct. Just please make sure CookieBanner will appear on every page that one can visit using URL. 
So it will appear also on pages like: 

https://localhost:3443/owner/0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266?network=hardhat

https://localhost:3443/vaults/list?network=hardhat

also ones that allow one to view somebody's vault in read only mode like:
https://staging.oasis.app/25862

Not sure how pages like /about but guess there it should appear too please verify with @FBrinkkemper 

https://staging.oasis.app/about

features/landing/CookieBanner.tsx
ferni: You made me realize that also a more accurate name for cookieSettings is "selectedCookies", that way one gets information about what the true/false value stands for, so I changed that as well.

ferni:
The build fails now after merging dev, but you can test it locally by undoing the merge commit: 57305208b452e8dfadeecef8e914e73b5db40f44

johnnieskywalker:
CookieBanner still doesn't appear on vault page
example URL https://localhost:3443/owner/0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266?network=hardhat

image

johnnieskywalker (approved):
Approved sir! Everything looks okay tested couple of scenarios, visiting vault, about team page and cookieBAnner appears everywhere

PR #595, opened 2021-11-18
Feature/four tiles

adamskrodzki:
Should implement a switcher between history/Overview and Protection, with currently only Overview implemented


components/TabSwitch.tsx
```diff
@@ -0,0 +1,44 @@
...
+}): JSX.Element {
+  const [mode, setMode] = useState<VaultViewMode>(defaultMode)
+
+  function handleToggle(newMode: VaultViewMode) {
+    setMode(newMode)
+  }
+  return (
+    <Grid gap={0} sx={{ width: '100%' }}>
+      <Grid columns={3} sx={{ zIndex: 1 }} variant="vaultEditingControllerContainer">
+        <Button onClick={() => handleToggle(VaultViewMode.Overview)}>Overview</Button>
```
johnnieskywalker, line 29: IMHO this are not 3 separate buttons but rather a component like this one
image

that has 3 states but styled in a different way that selected option is black

components/TabSwitch.tsx
```diff
@@ -0,0 +1,44 @@
...
+  protectionControl?: JSX.Element
+}): JSX.Element {
+  const [mode, setMode] = useState<VaultViewMode>(defaultMode)
+
+  function handleToggle(newMode: VaultViewMode) {
+    setMode(newMode)
+  }
+  return (
+    <Grid gap={0} sx={{ width: '100%' }}>
+      <Grid columns={3} sx={{ zIndex: 1 }} variant="vaultEditingControllerContainer">
```
johnnieskywalker, line 28: you can use it like here https://github.com/OasisDEX/oasis-borrow/blob/feature/four_tiles/features/manageVault/ManageVaultFormHeader.tsx#L51

PR #640, opened 2021-12-08
add ens name support

jelilat:
Integrate ENS name with the app frontend


adamskrodzki (changes requested):
@jelilat thanks for the PR, but it needs some work, 

first, since we support more networks we do not hardcode JsonRpcProvider creation in controls, it is available via context$

similarily additional information like ensName should be added to accountData$ and used from there

Thanks again for feature suggestion.  It is great to know that there are community members out there who care enough to make a PR to out codebase.

**I can adjust your PR myself later, or if you feel you chabe time and want to familiarise yoursel with our codebase You may try to do it yourself**

adamskrodzki:
@jelilat  closing for now, but based on Your PR we will add support after new year

PR #641, opened 2021-12-08
[technical debt] Presenters idea

sennett:
Rather than pass down a view model to the entire component tree, we can use presenters to hook into the business logic:

image

Current implementation uses the view data streams directly, but could also hook into different streams depending on the use-case when component is mounted.  Presenter would then take the different teams and map them into the components' props, and handle the clicks.

Advantages:

- not passing props down through everything (although AppContext is still used)
- provides an indirection layer between the business logic and the views (although this has already been provided by the streams themselves I think)
- possibly can think about reusing layouts, although exactly how is not clear to me at this point

Disadvantages:

- more code to write with this approach


features/manageVault/ManageVaultView.tsx
```diff
@@ -125,6 +129,21 @@ function ManageVaultForm(props: ManageVaultState) {
   )
 }
 
+const vaultHeaderPresenter = memoize(
```
sennett, line 132: This is what it looks like to use a presenter here.  Memoizing function avoids an infinite loop.  I'm currently looking at extracting this into an HOC or wrapper component.

features/manageVault/ManageVaultView.tsx
```diff
@@ -152,9 +171,11 @@ export function ManageVaultContainer({
     }
   }, [])
 
+  const vaultHeaderProps = usePresenter(useAppContext().manageVault$(id), vaultHeaderPresenter)
+  if (!vaultHeaderProps) return null
   return (
     <>
-      <DefaultVaultHeader {...manageVault} header={t('vault.header', { ilk, id })} id={id} />
+      <DefaultVaultHeader {...vaultHeaderProps} header={t('vault.header', { ilk, id })} />
```
sennett, line 178: This is what it looks like to use a presenter in a component. I'm looking at extracting this into an HOC or wrapper component.

features/manageVault/ManageVaultView.tsx
```diff
@@ -125,6 +129,21 @@ function ManageVaultForm(props: ManageVaultState) {
...
+          header: 'unused',
+          liquidationRatio: manageVaultState.ilkData.liquidationRatio,
+          stabilityFee: manageVaultState.ilkData.stabilityFee,
+          liquidationPenalty: manageVaultState.ilkData.liquidationPenalty,
+          debtFloor: manageVaultState.ilkData.debtFloor,
+        }
+      }),
+    ),
+)
+
```
sennett, line 146: This is what a present looks like.  They are created or contained in the component above that needs the props.  `memoize` was needed to avoid an infinite rendering loop as the observable is recreated each time.

helpers/usePresenter.ts
```diff
@@ -0,0 +1,15 @@
...
+import { useObservable } from './observableHook'
+
+export function usePresenter<T, U>(
+  getSources: (appContext: AppContext) => T,
+  createPresenter: ((sources: T) => Observable<U>) & MemoizedFunction,
+): U | undefined {
+  const sources: T = getSources(useAppContext())
+  const viewData$: Observable<U> = createPresenter(sources)
+  return useObservable(viewData$)
+}
```
piekczyk, line 15: General remark - the concept you provided is great because in that way we are able to prevent of props drilling. Nevertheless I'm not sure whether usage of whole appContext is necessary - because of that we will need to write presenters in the form of pipes instead of simple selectors. I though that since each product we have has its own state (which basically reflects app state at specific point of time) we could put this state inside new context and create hook which will allow to take slices of it by using selectors functions.

components/vault/VaultDetails.tsx
```diff
@@ -20,7 +21,7 @@ export type AfterPillProps = {
 }
 
 export function getCollRatioColor(
-  { inputAmountsEmpty, ilkData }: CommonVaultState,
+  { inputAmountsEmpty, ilkData }: { inputAmountsEmpty: boolean; ilkData: IlkData },
```
adamskrodzki, line 24: this is unrelated right ?

adamskrodzki (approved):
General remark - great improvement and possible to do gradually component after component.

helpers/usePresenter.ts
piekczyk: I've prepared a sketch how it could look like: 
https://github.com/OasisDEX/oasis-borrow/pull/646

features/manageVault/ManageVaultDetails.tsx
```diff
@@ -169,7 +169,12 @@ export function ManageVaultDetails(props: ManageVaultState) {
           afterPillColors={afterPillColors}
         />
 
-        <VaultDetailsCardCurrentPrice {...props} />
+        <VaultDetailsCardCurrentPrice
```
adamskrodzki, line 172: this is refactoring independent from presenter patterrn right ?

sennett: Yeah it was.

PR #666, opened 2021-12-14
move borrow open flow to selectors

sennett:
Goal here was to stop passing around the {...props} state through the tree, so that:

- we don't rely on a large state object being passed around to every component - nicer interfaces between components
- we don't need to re-render all components when something unrelated changes. (Although we currently still do because context fires the selectors.  Here's a lib that seems to solve this problem: https://github.com/dai-shi/use-context-selector, or another option is `memo` which we could use in the leaves of the react tree to prevent rerenders (https://reactjs.org/docs/react-api.html#reactmemo) - question here is whether we are passing down mutable or immutable objects from the streams into React - if we can always guarantee they are immutable, then perhaps `memo` on the leaves is enough).

File structure looks like this (further comments below huge image):

image

Context is created in `OpenVaultView`.  Almost every component ended up reading from the context, which in my mind is not ideal.  You can see the beginnings of a layout component in OpenVaultView -> OpenVaultContainer.


features/openVault/components/OpenVaultButton.tsx
```diff
@@ -83,7 +87,29 @@ export function OpenVaultButton(props: OpenVaultState) {
     token,
     depositAmount,
     generateAmount,
-  } = props
+    proxyAddress,
+    insufficientAllowance,
+    inputAmountsEmpty,
+    customAllowanceAmountEmpty,
+  } = useSelectFromContext(OpenBorrowVaultContext, (ctx) => ({
```
piekczyk, line 94: I was thinking about creating a hook per specific flow so it won't be necessary to pass Context every time you want to use state, but this approach is also fine since we won't have this X number of hooks which basically does the same but has different interface definition.

helpers/pick.ts
```diff
@@ -0,0 +1,8 @@
+// typesafe pick from https://stackoverflow.com/a/47232883/614523
+export function pick<T, K extends keyof T>(obj: T, ...keys: K[]): Pick<T, K> {
```
piekczyk, line 2: hm, I think ramda exports this method :)

https://ramdajs.com/docs/#pick

features/openVault/components/OpenVaultButton.tsx
sennett: Yeah I think I agree - better to use a small number of files.

helpers/pick.ts
sennett: Ah yes it does, but it's not typesafe:

```
import {pick} from "ramda";

function pickT<T, K extends keyof T>(obj: T, ...keys: K[]): Pick<T, K> {
    const ret: any = {}
    keys.forEach((key) => {
        ret[key] = obj[key]
    })
    return ret
}

// compiles despite missingKey not being present on object
pick(['presentKey', 'missingKey'], {presentKey: 'hello'})

// TS2345: Argument of type '"missingKey"' is not assignable to parameter of type '"presentKey"'.
pickT({presentKey: 'hello'}, 'presentKey', 'missingKey')
```

components/vault/VaultChangesInformation.tsx
```diff
@@ -58,7 +58,8 @@ export function VaultChangesInformationArrow() {
 }
 
 export function getEstimatedGasFeeText(
-  { gasEstimationStatus, gasEstimationUsd }: CommonVaultState,
+  gasEstimationStatus: GasEstimationStatus,
+  gasEstimationUsd: BigNumber | undefined,
```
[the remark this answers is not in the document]
sennett: Sorted.

features/openVault/components/Header.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+  const { headerProps } = useSelectFromContext(OpenBorrowVaultContext, (ctx) => ({
+    headerProps: {
+      ilkData: ctx.ilkData,
+      id: ctx.id,
+      header: t('vault.open-vault', { ilk: ctx.ilk }),
+    },
+  }))
+
+  return <DefaultVaultHeader {...headerProps} />
```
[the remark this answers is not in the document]
sennett: Sorted.

features/openVault/components/OpenVaultButton.tsx
```diff
@@ -1,23 +1,27 @@
...
-  insufficientAllowance,
-  inputAmountsEmpty,
-  customAllowanceAmountEmpty,
-}: OpenVaultState) {
+import { OpenVaultStage } from '../openVault'
+import { OpenBorrowVaultContext } from './OpenVaultView'
+
+function openVaultPrimaryButtonText(
+  stage: OpenVaultStage,
+  id: BigNumber | undefined,
```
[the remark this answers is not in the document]
sennett: Sorted.

PR #674, opened 2021-12-18
remove a duplicate view

sennett:
Applies some refactoring form Sonarcloud (nice one Adam), with goal of reducing duplication.

- Creates a common base type for defining vault stages `features/openMultiplyVault/common/TBaseManageVaultStage.ts`.
- Creates a new type `TAllVaultStages` which can be used by the view components so that they can accept all of the vault stages - i.e. the same view component used in different user flows.

From a recent discussion with Seba, `features/openMultiplyVault/common` is the place where we are starting to collect common things across these flows, but it currently contains files that should be renamed.  `features/openMultiplyVault` should be renamed to `features/vault` at some point in the future.  We should talk more to align on some form of file structure.


features/openMultiplyVault/common/TBaseManageVaultStage.ts
```diff
@@ -0,0 +1,21 @@
...
+  | 'collateralAllowanceWaitingForConfirmation'
+  | 'collateralAllowanceWaitingForApproval'
+  | 'collateralAllowanceInProgress'
+  | 'collateralAllowanceFailure'
+  | 'collateralAllowanceSuccess'
+  | 'daiAllowanceWaitingForConfirmation'
+  | 'daiAllowanceWaitingForApproval'
+  | 'daiAllowanceInProgress'
+  | 'daiAllowanceFailure'
+  | 'daiAllowanceSuccess'
```
paszkowskiDamian, line 16: Did we have dai Allowances in open multiply flow?

sennett: It is in both:

- [manage multiply vault](https://github.com/OasisDEX/oasis-borrow/pull/674/files#diff-9a0a1eddaa66411209f121193ab1b410ba34af9ec93a3afe8ce7b0df4c820d09L129-L133)
- [manage vault](https://github.com/OasisDEX/oasis-borrow/pull/674/files#diff-1e0d6743dffaa9fabbe32e5dc62e86c6eb35c274bf1e85f64570a13545f337aeL113-L117)

features/openMultiplyVault/common/ManageVaultCollateralAllowance.tsx
```diff
@@ -8,19 +9,32 @@ import React from 'react'
...
+  setCollateralAllowanceAmountToDepositAmount?: () => void
+  resetCollateralAllowanceAmount?: () => void
+  selectedCollateralAllowanceRadio: 'unlimited' | 'depositAmount' | 'custom'
+}
 
 export function ManageVaultCollateralAllowance({
   stage,
   vault: { token },
-  collateralAllowanceAmount,
   depositAmount,
```
piekczyk, line 29: depositAmount = zero

features/openMultiplyVault/common/ManageVaultCollateralAllowance.tsx
```diff
@@ -8,19 +9,32 @@ import React from 'react'
...
+  collateralAllowanceAmount,
   updateCollateralAllowanceAmount,
   setCollateralAllowanceAmountUnlimited,
   setCollateralAllowanceAmountToDepositAmount,
   resetCollateralAllowanceAmount,
   selectedCollateralAllowanceRadio,
-}: ManageVaultState) {
+}: ManageVaultCollateralAllowanceProps) {
+  const _depositAmount = depositAmount || new BigNumber(0)
```
piekczyk, line 37: no need if default is set for this variable

btw. instead of creating new big number which equals to 0 you can use "zero" from helpers

features/openMultiplyVault/common/ManageVaultCollateralAllowance.tsx
```diff
@@ -8,19 +9,32 @@ import React from 'react'
...
+type ManageVaultCollateralAllowanceProps = {
+  stage: TAllVaultStages
+  vault: { token: string }
+  depositAmount?: BigNumber
+  collateralAllowanceAmount?: BigNumber
+  updateCollateralAllowanceAmount?: (amount?: BigNumber) => void
+  setCollateralAllowanceAmountUnlimited?: () => void
+  setCollateralAllowanceAmountToDepositAmount?: () => void
+  resetCollateralAllowanceAmount?: () => void
+  selectedCollateralAllowanceRadio: 'unlimited' | 'depositAmount' | 'custom'
```
piekczyk, line 23: I believe that this type could be extracted as well

features/openMultiplyVault/common/ManageVaultCollateralAllowance.tsx
```diff
@@ -48,7 +62,7 @@ export function ManageVaultCollateralAllowance({
             onChange={setCollateralAllowanceAmountToDepositAmount!}
           >
             <Text variant="paragraph3" sx={{ fontWeight: 'semiBold', my: '18px' }}>
-              {t('token-depositing', { token, amount: formatCryptoBalance(depositAmount!) })}
+              {t('token-depositing', { token, amount: formatCryptoBalance(_depositAmount!) })}
```
piekczyk, line 65: ! - wont be necessary if default is defined

features/openMultiplyVault/common/TAllVaultStages.ts
```diff
@@ -0,0 +1,4 @@
+import { ManageMultiplyVaultStage } from '../../manageMultiplyVault/manageMultiplyVault'
+import { ManageBorrowVaultStage } from '../../manageVault/manageVault'
+
+export type TAllVaultStages = ManageMultiplyVaultStage | ManageBorrowVaultStage
```
piekczyk, line 4: This variable is called TAllVaultStages but contains only common state across managing flows. 

Additionally I'm not sure if we want to introduce new naming convention for types (I'm referring to "T" prefix).

Last thing - I know that current folder structure will be refactored soon but we could already think about folders devoted for types for specific flows.

piekczyk (changes requested):
Don't forget to delete this file as is empty for now :) 

features/openMultiplyVault/common/TProxyChange.ts

features/openMultiplyVault/common/ManageVaultCollateralAllowance.tsx
sennett: Nice.  Sorted.

features/openMultiplyVault/common/TAllVaultStages.ts
sennett: Yeah looking for feedback on the TPrefix.  I started adding it because this is a convention we used for Interfaces in C# which was helpful.  You can tell at a glance whether a file contains a Type (or Interface) or a concrete implementation.

Yeah I agree regarding folder structure.  We should do it in a quick fashion asap after christmas to avoid hilarious merge conflicts.

features/openMultiplyVault/common/ManageVaultCollateralAllowance.tsx
sennett: Sorted

features/openMultiplyVault/common/ManageVaultCollateralAllowance.tsx
sennett: Thanks/

features/openMultiplyVault/common/TAllVaultStages.ts
sennett: I removed the T prefix and put them in their own directory.

sennett:
> Don't forget to delete this file as is empty for now :)
> 
> features/openMultiplyVault/common/TProxyChange.ts

Done!

features/openMultiplyVault/common/ManageVaultDaiAllowance.tsx
```diff
@@ -8,18 +9,31 @@ import React from 'react'
...
+  daiAllowanceAmount?: BigNumber
+  updateDaiAllowanceAmount?: (amount?: BigNumber) => void
+  setDaiAllowanceAmountUnlimited?: () => void
+  setDaiAllowanceAmountToPaybackAmount?: () => void
+  resetDaiAllowanceAmount?: () => void
+  selectedDaiAllowanceRadio: SelectedDaiAllowanceRadio
+  paybackAmount?: BigNumber
+}
+
+export type SelectedDaiAllowanceRadio = 'unlimited' | 'paybackAmount' | 'custom'
```
piekczyk, line 25: not for now but we could think about moving types which are exported from components to common file (that should ensure that we won't fall into circular dependency)

.gitignore
```diff
@@ -21,7 +21,7 @@ public/sw.js.map
 public/workbox-*.js.map
 public/worker-*.js.map
 public/sitemap.xml
-types
+/types
```
piekczyk, line 24: hmm, normally this slash is not needed for defining paths in root folder. Since you are using macOs it's weird that you had to add it

piekczyk (approved):

features/openMultiplyVault/common/ManageVaultDaiAllowance.tsx
sennett: +1

.gitignore
sennett: It was needed because this rule was picking up the nested `features/openMultiplyVault/common/types` dir.

PR #697, opened 2022-01-12
Make borrow container optional in dev script.

ferni:
`./scripts/dev.sh` now launches the database without the borrow container.
`./scripts/dev-with-staging.sh` would launch it with the container, as before.


rkrasiuk (changes requested):
please add new empty line at the end of each file

adamskrodzki (dismissed):

rkrasiuk (approved):
great stuff

PR #698, opened 2022-01-12
FIX: Rounding Issues

rkrasiuk:

## Changes 👷‍♀️
- item 1
  
## How to test 🧪
- step 1 ...
    
## Definition of done ✔️


blockchain/utils.ts
```diff
@@ -6,13 +6,11 @@ import ethAbi, { AbiCoder } from 'web3-eth-abi'
 import { getToken } from './tokensMetadata'
 
 export function amountFromRay(amount: BigNumber): BigNumber {
-  BigNumber.config({ DECIMAL_PLACES: RAY.toString().length })
-  return amount.div(RAY)
```
rkrasiuk, line 10: this is bad practice because it changes the global configuration for `BigNumber`

adamskrodzki (changes requested):
have You tested Application for all this cases where You replaced amountToWei to amountToWeiRoundDown?

**I'm quite sure you did not**

and also I'm quite sure it is not a right approach in all cases. 

In your particular case problem I guess was that due to rounding up user tried to withdraw from vault more than he had (1 wei more probably) and that reverted transaction 

but in your 'fixes' there are also another cases like for instance paying back debt, 
and due to dust limit rounding down while paying debt will cause code to break because 
there can not 

General approach for the future:

1. Fix only the thing that is an issue 
2. Test every change You have made

You have changed literally every single funds transfering blockchain operation in that PR

adamskrodzki (changes requested):

PR #707, opened 2022-01-17
Ades/asset langing pages home page

sennett:
# [Create homepage](https://app.shortcut.com/oazo-apps/story/3228/create-homepage)

Functional changes:
- updates menu bar
- adds new homepage with product cards

Tech things:
- first pass at generic tab switcher that works in this context
- Creates generic product card container to hold and wrap lists of cards
- new link variant `inText` that takes its styling from the enclosing font.

Doesn't yet include info cards at the bottom of the homepage

## To test

- From homepage, create, edit, and close both multiply and borrow vaults.
- Do this with feature AssetLandingPages enabled and disabled.
- Test out homepage flow on mobile.
- Make sure the links in the headers point to the right places.


theme/index.tsx
```diff
@@ -589,6 +618,10 @@ const oasisBaseTheme = {
     },
   },
   links: {
+    inText: {
+      textDecoration: 'none',
+      color: '#585CF5',
```
piekczyk, line 623: maybe it is worth to add this color to colors in theme

piekczyk (approved):

PR #794, opened 2022-02-07
Fixed closing vault action when debt is zero

piekczyk:
# [Fixed closing vault action when debt is zero](https://clubhouseLink)

## Changes 👷‍♀️
- fixed closing vault action when debt is zero
  
## How to test 🧪
- heroku using multiply vault with and without debt
    
## Definition of done ✔️

### DIFF
--- a/features/multiply/manage/pipes/manageMultiplyVaultTransactions.ts
+++ b/features/multiply/manage/pipes/manageMultiplyVaultTransactions.ts
@@ -663,7 +663,7 @@ export function applyEstimateGas(
           ilk,
         })
       } else {
-        if (state.otherAction === 'closeVault' && !state.vault.debt.isZero()) {
+        if (state.otherAction === 'closeVault' && !debt.isZero()) {
           const { fromTokenAmount, toTokenAmount, minToTokenAmount } =
             closeVaultTo === 'dai' ? closeToDaiParams : closeToCollateralParams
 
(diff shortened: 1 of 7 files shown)


features/multiply/manage/pipes/manageMultiplyVaultTransactions.ts
```diff
@@ -663,7 +663,7 @@ export function applyEstimateGas(
           ilk,
         })
       } else {
-        if (state.otherAction === 'closeVault') {
+        if (state.otherAction === 'closeVault' && !state.vault.debt.isZero()) {
```
peculiarity, line 666: You already have `debt` as variable destructured. Used it directly.

features/multiply/manage/pipes/manageMultiplyVaultTransitions.ts
```diff
@@ -266,6 +266,9 @@ export function progressAdjust(
     if (state.withdrawAmount !== undefined || state.paybackAmount !== undefined) {
       return manageVaultWithdrawAndPayback(txHelpers$, change, state)
     }
+    if (state.otherAction === 'closeVault' && state.vault.debt.isZero()) {
```
peculiarity, line 269: Um I might be wrong but if you put everywhere this condition to  hide `closeVault` if the debt is zero, how do we get into the situation that the otherAction is `closeVault` and the `debt` is also zero?

piekczyk: Here I just added another case of possible progress action. It's more like switching than  hiding. Close vault action will be always visible on UI. The only change is that this actions has underneath two cases:
1) normal close vault action (close to dai / close to eth)
2) close action when debt = 0

PR #814, opened 2022-02-11
Read from DSS Charter proxy actions contract

sennett:
Creates the ability to call new proxy actions contract for institutional vaults including the ABIs - should just need to configure the the charter proxy actions contract address and inject the different contract to start reading from it 🤞

I created a generic interface for DSS Proxy smart contracts to avoid a bunch of ifs in what looks like business logic of `getWithdrawAndPaybackCallData` and `getDepositAndGenerateCallData`, and so that we can switch them out at the top of `createManageVault$` without affecting anything else in the pipe.  This seems like business logic though - is there a better place for it?

Testing:

- open + manage borrow + multiply vaults


blockchain/calls/proxyActions.ts
```diff
@@ -31,98 +33,80 @@ export type WithdrawAndPaybackData = {
...
-    return contract<DsProxy>(contractDesc(dsProxy, proxyAddress)).methods['execute(address,bytes)']
-  },
-  prepareArgs: (data, context) => {
-    const { dssProxyActions } = context
-    return [dssProxyActions.address, getWithdrawAndPaybackCallData(data, context).encodeABI()]
-  },
+export interface IProxyActions {
+  withdrawAndPayback: TransactionDef<WithdrawAndPaybackData>
+  depositAndGenerate: TransactionDef<DepositAndGenerateData>
+}
```
sennett, line 75: I don't love the name of this.  Any ideas?

blockchain/calls/proxyActions/DssProxyActionsSmartContractWrapperInterface.ts
```diff
@@ -0,0 +1,75 @@
+import {
+  NonPayableTransactionObject,
+  PayableTransactionObject,
+} from '../../../types/web3-v1-contracts/types'
+import { ContextConnected } from '../../network'
+import { DepositAndGenerateData, WithdrawAndPaybackData } from './proxyActions'
+
+export interface DssProxyActionsSmartContractWrapperInterface {
+  resolveContractAddress: (context: ContextConnected) => string
```
paszkowskiDamian, line 9: beautiful 🤟 ❤️

paszkowskiDamian (approved):

PR #822, opened 2022-02-15
First-pass sentry integration

sennett:
Adds [Sentry](https://sentry.io/) to oasis-borrow.  Sentry gives us the ability to see client-side errors in production, and more besides but that may require more integration.

- Creates things in Sentry:
  - Environments: `development` for local machine, `pullrequest` for PR dynos.  Staging and production TBC, but likely `staging` and `production`.
  - Releases: `development` for local machine, `$HEROKU_APP_NAME` for PR dynos.  Commit hash for staging and production.
- New local file `.sentryclirc` required, which contains secret auth token to upload source maps.  Configured in env vars in higher enviroments.
- New environment variables:
  - `SENTRY_AUTH_TOKEN` - allows build process to upload source maps to sentry.
  - `NEXT_PUBLIC_SENTRY_ENV` - used to specify the environment of the events in Sentry so we can filter - i.e. `development`, `pullrequest`, `production` etc.
  - `SENTRY_RELEASE` - used to specify the release in Sentry:
    - Local dev: `development`
    - Heroku: `pullrequest`
    - Staging and production: $SOMMIT_SHA
- `/errors` page to trigger errors and test the integration with Sentry.

Testing:
---

I've tested it with both heroku envs and locally, and things appear to be working (i.e. we can see errors and stack traces in sentry).  I'm going to run this in prod for a week or so, see how it goes, and iron out any kinks.

There is another file that contains an auth token needed to actually send sourcemaps to sentry when you build (doesn't happen on dev builds).  Ask me if it errors out and I can share it.


.gitignore
```diff
@@ -11,6 +11,7 @@ yalc.lock
 .idea
 .log
 .DS_Store
+.sentryclirc
```
johnnieskywalker, line 14: [minor] Justa remark. We could also add .nvmrc to run `nvm use` each time you open project however to make it work properly we should set a newer version of node because now when one runs `nvm use` in borrow node version is set to 11

README.md
```diff
@@ -165,6 +169,10 @@ at _run time_
 
 - `MAILCHIMP_API_KEY` - Mailchimp API Key used to integrate Mailchimp newsletter.
 
+- `SENTRY_RELEASE` - The release in sentry.io.  Used by sentry.io to generate and upload source maps for a given release at biudl time, and tie those source maps to errors sent to sentry at run time.
```
johnnieskywalker, line 172: [minor] _given release at_ either build or build but not biudl ;)

johnnieskywalker (changes requested):

.gitignore
sennett: Yeah Damian is looking into the Node version I think.  Let's wait for him before doing that.

johnnieskywalker (approved):

pages/errors/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+      </li>
+      <li>
+        <a href="/api/deliberateError">trigger error on API endpoint</a>
+      </li>
+    </ul>
+  )
+}
+
+ServerError.layout = MarketingLayout
+ServerError.theme = 'Landing'
```
peculiarity, line 73: What's the point of this whole page?

sennett: To test the sentry integration - that errors are raised and tagged properly in different envs.  It will be removed soon.  I will make it only available on staging not prod, and remove from source maps - good spot.

components/errorTriggeringComponents/TriggerErrorWithUseObservableWithError.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { of } from 'rxjs'
+import { flatMap } from 'rxjs/operators'
+
+import { useObservableWithError } from '../../helpers/observableHook'
+const streamThatErrors$ = of(1).pipe(flatMap(() => fetch('https://fetch-handled-url')))
```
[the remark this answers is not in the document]
sennett: Hmm this doesn't seem to be triggering the error?

PR #843, opened 2022-02-21
Prepare manage borrow vault pipeline for insti vaults

sennett:
Builds off Damian's work and prepares manage borrow pipeline to support charter (institutional) vaults, and integrates into the view created by Ferni.

There's no user-facing change at present.  Technical notes:

- new `InstiVault$` that wraps standard Vault and reads from additional smart contract.
- `createManageVault$` is used for both `manageVault$` and `manageInstiVault$`.  `createManageVault$` receives initial view state and `apply` function as a dependency, and there is one instance of `BorrowManageVaultViewStateProviderInterface` for `manageVault$` and one for `manageInstiVault$`.
- `ManageVaultState` renamed to `ManageStandardBorrowVaultState`.
- provides a place in AppContext to hardcode vault IDs which we can use if we can't determine vault type (`Borrow | Insti`) from the smart contracts (perhaps DB is better to avoid this logic?)
- `features/borrow/manage/tests/manageVault.test.ts` looks a bit intense, but it's just repeating the same tests for `manageVault$` and `manageInstiVault$` so that the same user interactions work for both, and adding some other insti vault specific things.

Testing:

Can't currently test managing insti vaults as we don't have smart contracts deployed, but we can create a borrow vault and then flag it as an insti vault:

1. create a standard borrow vault.
2. mark it as an insti vault in `HARDCODED_VAULT_TYPES` in AppContext
3. Set the `withdrawPaybackDepositGenerateLogicFactory` to `StandardDssProxyActionsContractWrapper` on line 513-515.
4. When creating `instiVault$` in AppContext, use the dummy pipes that are commented out rather than `charterNib$, charterPeace$, charterUline$`
5. Refresh the page.
6. You should now see the insti vault UI and be able to interact with the vault in the normal way.

- Also need regression testing on manage borrow vault before deploy.

Next steps (future PR):
- hydrate InstiVault termEnd, fixedFee, and nextFixedFee when we get the new smart contracts that fullfill these.
- determine institutional vault types either with a smart contract call or by populating HARDCODED_VAULT_TYPES


blockchain/instiVault.test.ts
```diff
@@ -0,0 +1,90 @@
...
+          usr: string | undefined
+        }
+      | undefined
+
+    const uncalledCharterPipeArgs: CharterPipeArgs = {
+      ilk: 'no',
+      usr: 'no',
+    }
+
+    let createCharterNib$FakeCalledWith: CharterPipeArgs = uncalledCharterPipeArgs
```
paszkowskiDamian, line 54: `$` means that variable is a Observable, so I guess here it is not necessary. This is supposed to be a Hungarian notation and also rxjs convention.

blockchain/config.ts
```diff
@@ -253,6 +256,7 @@ const goerli: NetworkConfig = {
   mcdEnd: contractDesc(mcdEnd, goerliAddresses.MCD_END),
   mcdSpot: contractDesc(mcdSpot, goerliAddresses.MCD_SPOT),
   mcdDog: contractDesc(mcdDog, goerliAddresses.MCD_DOG),
+  dssCharter: contractDesc(dssCharter, '0x0000'),
```
paszkowskiDamian, line 259: We can update this with proper address.

We can do it in next PR

blockchain/instiVault.ts
```diff
@@ -0,0 +1,53 @@
...
+              // todo: insti-vault put real values here
+              ...vault,
+              originationFeePercent: nib,
+              activeCollRatio: peace,
+              activeCollRatioPriceUSD: peace,
+              debtCeiling: uline,
+              termEnd: new Date('2022-03-02'),
+              fixedFee: new BigNumber(0.2),
+              nextFixedFee: new BigNumber(0.3),
+              instiIlkName: 'INST-NEXO-ETH',
```
paszkowskiDamian, line 46: I think it will be just ilk name, but we will check it.

components/AppContext.ts
```diff
@@ -343,6 +358,17 @@ export function setupAppContext() {
     bigNumberTostring,
   )
 
+  const instiVault$ = memoize(
+    // todo: insti-vault switch back to smart contract vaules when contract is deployed
+    curry(createInstiVault$)(vault$, charterNib$, charterPeace$, charterUline$),
+    // curry(createInstiVault$)(
```
paszkowskiDamian, line 364: I guess comment is not necessary any more.

components/AppContext.ts
```diff
@@ -534,9 +583,17 @@ export function setupAppContext() {
...
   )
 
-  const checkVault$ = memoize((id: BigNumber) => curry(checkVaultTypeUsingApi$)(context$, id))
+  // const HARDCODED_VAULT_TYPES: VaultIdToTypeMapping = { 27609: VaultType.Insti }
+  const HARDCODED_VAULT_TYPES: VaultIdToTypeMapping = {}
+
+  const checkVault$: (id: BigNumber) => Observable<VaultType> = curry(createCheckVaultType$)(
+    curry(checkVaultTypeUsingApi$)(context$),
+    HARDCODED_VAULT_TYPES,
+  )
```
paszkowskiDamian, line 592: maybe we can distinguish by ilk 🤔 but need to check contracts

paszkowskiDamian (dismissed):

components/AppContext.ts
sennett: Yeah we can switch now.  I'll do it when we moved back to insti vaults.

components/AppContext.ts
sennett: That would be nice.  Or we can use vault type in DB.

paszkowskiDamian (approved):

PR #849, opened 2022-02-24
raise observable errors in sentry

sennett:
Makes it so we can see observable errors in sentry:

image

Also removes `useObservable` hook and renames `useObservableWithError` -> `useObservable`.

Also removes test pages that were used to test sentry errors.


components/vault/HistoryControl.tsx
```diff
@@ -15,10 +15,8 @@ interface HistoryControlProps {
 
 export function HistoryControl({ generalManageVault }: HistoryControlProps) {
   const { vaultHistory$, vaultMultiplyHistory$ } = useAppContext()
-  const vaultHistoryWithError = useObservableWithError(
-    vaultHistory$(generalManageVault.state.vault.id),
-  )
-  const vaultMultiplyHistoryWithError = useObservableWithError(
+  const vaultHistoryWithError = useObservable(vaultHistory$(generalManageVault.state.vault.id))
```
peculiarity, line 18: I would rename these to `vaultHistory` and `vaultMultiplyHistory`

components/vault/ProtectionControl.tsx
```diff
@@ -45,8 +45,8 @@ interface ProtectionControlProps {
 export function ProtectionControl({ vault, ilkData, account }: ProtectionControlProps) {
   const { automationTriggersData$, collateralPrices$ } = useAppContext()
   const autoTriggersData$ = automationTriggersData$(vault.id)
-  const automationTriggersDataWithError = useObservableWithError(autoTriggersData$)
-  const collateralPricesWithError = useObservableWithError(collateralPrices$)
+  const automationTriggersDataWithError = useObservable(autoTriggersData$)
+  const collateralPricesWithError = useObservable(collateralPrices$)
```
peculiarity, line 49: Probably we can remove the `error` bit in here as well

components/vault/ProtectionControl.tsx
```diff
@@ -45,15 +45,13 @@ interface ProtectionControlProps {
...
-    <WithErrorHandler
-      error={[automationTriggersDataWithError.error, collateralPricesWithError.error]}
-    >
+    <WithErrorHandler error={[automationTriggersData.error, collateralPrices.error]}>
       <WithLoadingIndicator
-        value={[automationTriggersDataWithError.value, collateralPricesWithError.value]}
+        value={[automationTriggersData.value, collateralPrices.value]}
         customLoader={<VaultContainerSpinner />}
       >
         {([automationTriggersData, collateralPrices]) => {
```
piekczyk, line 57: doesn't lint screams here that automationTriggersData has been already declared somewhere?

components/vault/ProtectionControl.tsx
sennett: Done.

components/vault/HistoryControl.tsx
sennett: Done

components/vault/ProtectionControl.tsx
sennett: Doesn't seem to no - lint doesn't seem to care about using the same variable name inside the closure.

sennett: After some investigation it seems we don't have 'no-shadow' rule set.  This is what happens when we enable it for eslint 😁

image

ferni (dismissed):

piekczyk (approved):

peculiarity (approved):

PR #875, opened 2022-03-09
Uniswap widget

ferni:
# [Uniswap widget](https://app.shortcut.com/oazo-apps/story/3618/uniswap-widget)

## Changes 👷‍♀️
 
- Add an exchange button at the top which opens the widget.
- It's disabled on mobile.
  
## How to test 🧪

- Enable the feature on your browser console: `localStorage.setItem('features', '{"Exchange":true}')`
- On goerli it just has two tokens, on mainnet you would see all the chosen tokens.


components/Header.tsx
```diff
@@ -97,6 +98,12 @@ interface UserAccountProps {
 function UserAccount({ position }: UserAccountProps) {
   const { vaultFormToggleTitle, setVaultFormOpened } = useSharedUI()
 
+  const web3Provider = (() => {
```
peculiarity, line 101: What's the point of having self invoking function ?

components/Header.tsx
```diff
@@ -112,6 +119,9 @@ function UserAccount({ position }: UserAccountProps) {
       }}
     >
       <Flex>
+        <Box sx={{ display: ['none', 'flex'] }}>
+          <ExchangeButton web3Provider={web3Provider} />
```
peculiarity, line 123: I guess if we don't have provider we shouldn't display that button or at least it should be disabled. What do you think ?

helpers/useOnClickOutside.ts
```diff
@@ -0,0 +1,15 @@
+import React, { useEffect } from 'react'
+
+export function useOnClickOutside(ref: React.RefObject<HTMLDivElement>, cb: Function) {
```
peculiarity, line 3: I would rename it to `handleOutsideElementClick`. What do you think?

components/Header.tsx
ferni: To limit the scope of the consts that are only used to produce the web3Provider one, reducing complexity.

components/Header.tsx
ferni: Sounds good.

helpers/useOnClickOutside.ts
ferni: I like the name change, but shouldn't hooks begin with "use"?

ferni:
I'm not sure what's the deal with the typecheck errors.

components/Header.tsx
peculiarity: Why would you want to limit the scope of the consts?

helpers/useOnClickOutside.ts
peculiarity: `useOutsideElementClickHandler`

components/Header.tsx
sennett: This web3 provider could be inside `ExchangeButton` I think to avoid passing it down?

components/exchangeMenu/ExchangeButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+  web3Provider?: provider
+  sx?: SxStyleProp
+}) {
+  const componentRef = useRef(null)
+  const [isOpen, setIsOpen] = useState(false)
+
+  useOnClickOutside(componentRef, () => setIsOpen(false))
+
+  return (
+    <Flex ref={componentRef} sx={{ position: 'relative', mr: 2, pr: 1, ...sx }}>
```
sennett, line 22: I love this hook you've made.  Could be like this and return the ref from the hook?

```suggestion
  const componentRef = useOnClickOutside(() => setIsOpen(false))
  const [isOpen, setIsOpen] = useState(false)

  return (
    <Flex ref={componentRef} sx={{ position: 'relative', mr: 2, pr: 1, ...sx }}>
```

components/Header.tsx
ferni: > Why would you want to limit the scope of the consts?

It just makes sense for me to write it that way, semantically. I only care about the web3Provider, and that is clearly expressed in the code.

If i wanted to, for instance, change how web3Provider is obtained, I could replace all the consts used inside the function with confidence,  knowing the aren't being used somewhere else.

> This web3 provider could be inside ExchangeButton I think to avoid passing it down?

Yeah I did it this way because I pass a different provider in Storybook, but I guess it could be an optional param.

helpers/useOnClickOutside.ts
ferni: 👍

components/exchangeMenu/ExchangeButton.tsx
ferni: Woah, nice, I'll try it.

ferni: Done.

helpers/useOutsideElementClickHandler.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useRef } from 'react'
+
+export function useOutsideElementClickHandler(cb: Function) {
```
sennett, line 3: Love this.

sennett (approved):

PR #893, opened 2022-03-16
disable steCRV button when user has a vault already

sennett:
Disables steCRV when the user already has a steCRV vault, because a wallet can only have one.

image

Creates a pipe `canCreateVaultForIlk$` which is consumed by `productCardData$` provides an excuse if the user cannot open the vault.

Testing:

- create a full vault where the debt ceiling is zero - ensure button disabled and shows as `full`
- ensure you can create a steCRV borrow position (both connected and disconnected wallet) if you don't already have one
- ensure you button is disabled when you go to create another steCRV borrow position

### DIFF
--- a/components/AppContext.ts
+++ b/components/AppContext.ts
@@ -53,7 +53,13 @@ import {
   createBalance$,
   createCollateralTokens$,
 } from 'blockchain/tokens'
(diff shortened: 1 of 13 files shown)


paszkowskiDamian:
One more thing, we should redirect to user to their vault from `/open/*` page.

PR #921, opened 2022-03-30
Add rollback GHA

jeannettemcd:
CICD to allow Devs to roll production back to a previously released version.
This is a **manual** trigger job, and requires the user to input the short commit hash, of a previously released tag.
The docker image will then be re-tagged as **_latest_**, and deployed to prod.
This way dev's can deploy any release tag that they choose, and not disrupt normal workflows, as all deployed images will be tagged with **_:latest_** .
cc: @peculiarity

### DIFF
--- a/.github/workflows/aws-prod-rollback.yaml
+++ b/.github/workflows/aws-prod-rollback.yaml
@@ -1,0 +1,78 @@
+name: Rollback Production
+
+on:
+  workflow_dispatch:
+    branches: [ main ]
+    inputs:
+      commit_hash:
+        description: 'Enter the commit hash you want to roll back to.'
+        default: '<hash>'
+        required: true
+        type: string
+
+jobs:
+  rollback-deployment:
+    env:
+      AWS_REGION: us-east-1
+      ECR_REPO_NAME: oasis-borrow-prod
+      SERVICE_NAME: oasis-borrow-prod
+      CLUSTER_NAME: oasis-borrow-prod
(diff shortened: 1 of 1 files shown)


cristidas:
@jeannettemcd being that we're using ECR, we could also use this approach, instead of pulling and pushing the image again:

```
MANIFEST=$(aws ecr batch-get-image --repository-name $ECR_REPO_NAME --image-ids imageTag=${{ env.COMMIT_HASH }} --query 'images[].imageManifest' --output text)
aws ecr put-image --repository-name $ECR_REPO_NAME --image-tag latest --image-manifest "$MANIFEST"
```

What do you think?

paszkowskiDamian:
How reverting to normal flow would look like?

jeannettemcd:
> 
This method of rollback preserves the normal workflow. 
Production always runs the docker image tagged as "latest". By retaging the image, you want to roll back to, as "latest", this compliments the normal workflow. In otherwords after a rollback, dev can create a fix, merge & deploy it to production as normal, because the workflow will build and deploy a new "latest" image. 
At build time, all images are tagged with two tags, the short commit hash of the release, and "latest". The image with the "latest" tag, is the one running in production, so by manipulating "latest" tag, we can roll back and forth without breaking any workflows, or making any infra changes.

jeannettemcd:
> 

I like it. Will test this out in test and let you know.

cristidas (approved):

eskp (approved):
Good stuff!

PR #922, opened 2022-03-30
Uniswap widget style

ferni:


package.json
```diff
@@ -213,9 +213,10 @@
     "patch-package": "^6.2.2",
     "prettier": "^2.0.1",
     "prisma": "^3.9.1",
+    "react-i18next": "^11.16.2",
```
sennett, line 216: Not sure we use this anymore?

sennett (approved):

package.json
ferni: I added this to make storybook work. Didn't manage to make it work because of some theme-ui thing though.

PR #958, opened 2022-04-14
Update vault history section

piotrkonowrocki:
# [Update vault history section](https://app.shortcut.com/oazo-apps/story/3860/redesign-update-vault-history-section)
  
## Changes 👷‍♀️

- Updated layout in vault history section.
- Removed extensive usage of `div`s all over the place and changed some of them into other tags according to `HTML` semantics.
  
## How to test 🧪

Check out manage vault page or, in case of automation feature flag enabled, "history" tab on vault page. Please not that you need to check multiply vault to see events on in history that has details.

### DIFF
--- a/features/vaultHistory/VaultHistoryEntry.tsx
+++ b/features/vaultHistory/VaultHistoryEntry.tsx
@@ -1,0 +1,333 @@
+import { Icon } from '@makerdao/dai-ui-icons'
+import BigNumber from 'bignumber.js'
+import { amountFromWei } from 'blockchain/utils'
+import { AppLink } from 'components/Links'
+import { WithArrow } from 'components/WithArrow'
+import {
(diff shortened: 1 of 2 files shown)


piekczyk:
Hey. This redesign changes should be hidden behind AutomationBasicBuyAndSell feature toggle. Feature toggle Automation and current production version should use old design.

piotrkonowrocki:
Okay, done, moved into `AutomationBasicBuyAndSell` but that required to copy old history components and update them separately. Conditioning every layout change in old component would lead to unreadable garbage.

Old components should be removed when `AutomationBasicBuyAndSell` flag is no longer needed, I left a comment in the code so the person who is going to do that would remember about it.

piekczyk (approved):

PR #961, opened 2022-04-15
Owner page vault loading time fix

piekczyk:
# [Owner page vault loading time fix](https://shortcutLink)

## Changes 👷‍♀️
- removed initial state from pipe that delivers data for owner page
  
## How to test 🧪
- goerli with automation flag turned on, you can use following owner 0xB710940E3659415ebd5492f43B247891De14D872
- compare it with staging instance and see if there is an loading speed improvement

### DIFF
--- a/features/vaultsOverview/vaultsOverview.ts
+++ b/features/vaultsOverview/vaultsOverview.ts
@@ -11,7 +11,7 @@ import { useFeatureToggle } from '../../helpers/useFeatureToggle'
 import { extractStopLossData } from '../automation/common/StopLossTriggerDataExtractor'
 import { TriggersData } from '../automation/triggers/AutomationTriggersData'
 import { ilksWithFilter$, IlksWithFilters } from '../ilks/ilksFilters'
-import { vaultsWithFilter$, VaultsWithFilters } from './vaultsFilters'
+import { vaultsWithFilter$, VaultsWithFilters, VaultWithSLData } from './vaultsFilters'
 import { getVaultsSummary, VaultSummary } from './vaultSummary'
 
 export interface VaultsOverview {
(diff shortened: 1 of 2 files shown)


features/vaultsOverview/vaultsOverview.ts
```diff
@@ -45,17 +45,19 @@ export function createVaultsOverview$(
...
 
+  const borrowVaults = ((automationEnabled ? vaultWithAutomationData$ : vaults$(address)).pipe(
+    map((vaults) => vaults.filter((vault) => vault.type === 'borrow')),
+    // TODO casting won't be necessary when Automation feature flag will be removed
+  ) as unknown) as Observable<VaultWithSLData>
+
+  const multiplyVaults = ((automationEnabled ? vaultWithAutomationData$ : vaults$(address)).pipe(
+    map((vaults) => vaults.filter((vault) => vault.type === 'multiply')),
+    // TODO casting won't be necessary when Automation feature flag will be removed
+  ) as unknown) as Observable<VaultWithSLData>
```
adamskrodzki, line 56: why not just 
` const allVaults = ((automationEnabled ? vaultWithAutomationData$ : vaults$(address)).pipe(
    map((vaults) => vaults.filter((vault) => ( vault.type === 'multiply' || vault.type ==='borrow') )),  ) as unknown) as Observable<VaultWithSLData>
`

adamskrodzki: and then have single vaultsWithFilter$

features/vaultsOverview/vaultsOverview.ts
```diff
@@ -45,17 +45,19 @@ export function createVaultsOverview$(
     }),
   )
 
+  const borrowVaults = ((automationEnabled ? vaultWithAutomationData$ : vaults$(address)).pipe(
+    map((vaults) => vaults.filter((vault) => vault.type === 'borrow')),
+    // TODO casting won't be necessary when Automation feature flag will be removed
+  ) as unknown) as Observable<VaultWithSLData>
+
+  const multiplyVaults = ((automationEnabled ? vaultWithAutomationData$ : vaults$(address)).pipe(
```
adamskrodzki, line 53: is this generic enough to not be changed everytime new kind of trigger is introduced? Why automation data are event relevant here ?

adamskrodzki (changes requested):

piotrkonowrocki (commented):
How big should be the difference in loading time between your branch and current main? I've tested it locally and sometimes I don't see a change at all, sometimes it can be faster by around 500ms.

But maybe it's because local environment isn't the fastest itself and it needs to be tested on staging?

features/vaultsOverview/vaultsOverview.ts
piekczyk: I guess it was done in that way because later borrow and multiply vaults are mapped as different properties so it's easier to split it on this level. I just extracted it as consts.

`  return combineLatest(
    vaultsWithFilter$(borrowVaults),
    vaultsWithFilter$(multiplyVaults),
    vaultsAddress$.pipe(map(getVaultsSummary)),
    ilksWithFilter$(ilksListWithBalances$),
  ).pipe(
    map(([borrow, multiply, vaultSummary, ilksWithFilters]) => ({
      vaults: {
        borrow,
        multiply,
      },
      vaultSummary,
      ilksWithFilters,
    })),
    distinctUntilChanged(isEqual),
  )`

features/vaultsOverview/vaultsOverview.ts
piekczyk: We will need to extend vaultWithAutomationData$ pipe (and rename it for sure) with all extra stuff that we need - currently only stop loss information is needed. It is relevant here because we are showing Stop Loss On labels on the owner page where the borrow / multiple vault lists are. 

image

piekczyk:
> How big should be the difference in loading time between your branch and current main? I've tested it locally and sometimes I don't see a change at all, sometimes it can be faster by around 500ms.
> 
> But maybe it's because local environment isn't the fastest itself and it needs to be tested on staging?

Main issue here was that owner page was loaded with empty vault lists and after some time list were loaded. Right now when loading spinner disappears all data should be already there (number of vaults + vault lists). With this fix whole state should be generated all at once instead of providing dummy initial state for lists and then waiting till new block arrives (so in the end it should be faster but it's hard to say how much faster since it depends on blocks).


Issue described above is visible on the screen below
image

adamskrodzki (approved):

piotrkonowrocki (approved):
Okay, now I see, works as intended 👌

PR #968, opened 2022-04-21
Vault info section as vault tab

piekczyk:
# [Vault info section as vault tab](https://app.shortcut.com/oazo-apps/story/3859/vault-information-section)

## Changes 👷‍♀️
- added new tab with vault info
- removed old vault info from header
  
## How to test 🧪
- enable feature toggle AutomationBasicBuyAndSell and Automation


piotrkonowrocki (commented):
I have two suggestions, they are not request changes, but I will wait with approval for your response.

1. Last week I made new version of vault history page. Single item in new layout looks the same as vault info that you just did. Maybe you could try extract insides of my vault history component as new components, something like `DefinitionList` and use it in both places? Link to PR with vault history: https://github.com/OasisDEX/oasis-borrow/pull/958.
2. This one concerns not just this PR but all future ones. I noticed that across entire oasis app there are rarely any semantics in html tags. Everything is just another div. Don't you think that we should use at least basic tags as `ul`, `li`, `p` etc.? Theme UI allows to do that with `as` prop that could be use on most of its components. I know that it's probably too late to fix old components, be maybe we could try keep that in mind for future ones. What do you think?

PR #970, opened 2022-04-21
Sl ratio validator fix

piekczyk:
# [Sl ratio validator fix](https://shortcutLink)

https://www.notion.so/oazo/Bugs-and-Suggestions-3fa8565c715f4c31bcdeb07cc8e9e76b

## Changes 👷‍♀️
- changed validation from `lower than` to `lower than or equal to`
  
## How to test 🧪
- setup stop loss to for example 200% and then try to modify vault to push it to coll ratio equal to 200% - in this case error should popup


features/form/commonValidators.ts
```diff
@@ -404,7 +404,7 @@ export function afterCollRatioBelowStopLossRatioValidator({
   stopLossRatio: BigNumber
 }) {
   return (
-    afterCollateralizationRatio.lt(stopLossRatio) ||
-    afterCollateralizationRatioAtNextPrice.lt(stopLossRatio)
+    afterCollateralizationRatio.lte(stopLossRatio) ||
```
adamskrodzki, line 407: shouldn't there be some margin applied ?

adamskrodzki (approved):

PR #976, opened 2022-04-22
chore: Update PIP_CRVV1ETHSTETH address for mainnet

peculiarity:
The value is taken from here https://github.com/makerdao/spells-mainnet/blob/PE-958/src/DssSpell.sol/


paszkowskiDamian:
should we also re-enable crv ilk? We had to remove cards navigating to opening this vault.

paszkowskiDamian (approved):

PR #979, opened 2022-04-26
Add smart contract address unit tests for OpenVault pipeline

BackToTheCode:
Adds unit tests that make assertions that contract descriptions (abi + address) on shared context are what we expect. Also, where possible (proxy + allowance) tests that the t/x objects created by sendWithGasEstimate conform to a t/x object created manually inside our tests.

In order to get confidence opted to mock as few of our modules as possible. This meant introducing additional mocks - MockProvider.mock.ts given a provider is expected by libraries that the '@oasisdex lib depends on.

**Next steps**
Write similar tests for manage-borrow/open-multiply/manage-multiply/guni.. etc pipelines

**Potential improvements**
Write integration tests that run tests against a hardhat node. With this we can test proxyAddresses are correct which is potentially the larger risk. Also, these tests could query Maker (and other's) chain logs or smart contract registries.


blockchain/calls/proxyActions/vaultActionsLogic.ts
```diff
@@ -156,3 +164,16 @@ function getOpenCallData(
   }
   return proxyActionAdapter.open(context, data)
 }
+
+export function mockVaultActionsLogicWithSpies(sinon: SinonStatic) {
```
peculiarity, line 168: I strongly disagree to have this code in production files. This is something that lives only in the scope of tests. You can define that somewhere there. These files that are sent to the user doesn't need to know about any mocks or sinon libraries or etc.

peculiarity (approved):

blockchain/ilks.test.ts
```diff
@@ -27,5 +27,6 @@ describe('ilkData$', () => {
     debtScalingFactor$.next(RANDOM_DEBT_SCALING_FACTOR)
     expect(state().ilkDebt.gt(defaultIlkDebt)).to.be.true
     expect(state().ilkDebtAvailable.lt(expectedIlkDebtAvailable)).to.be.true
+    done()
```
adamskrodzki, line 30: shouldnt function be async instead ? if everything is sync here then done should not be needed

helpers/mocks/web3Context.mock.ts
```diff
@@ -21,32 +23,42 @@ export const mockWeb3ContextConnectedReadonly: Web3ContextConnectedReadonly = {
...
+  account: '0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266',
+  proxy: '0xf7203265834862bA831bc14e8999C2D1dBa7A645',
+}
 
 export const mockWeb3ContextConnected: Web3ContextConnected = {
   status: 'connected',
   connectionKind: 'injected',
   web3: new Web3(),
-  chainId: 1,
+  chainId: kovanNetworkId,
```
adamskrodzki, line 37: kovan? It's decommisioned, it should be goerly if any

BackToTheCode:
Proposing we close this PR because doing this as unit-tests isn't really practical and we should really explore this using with integration tests.

PR #987, opened 2022-04-28
Piechart

ferni:
# [Create Piechart component](https://app.shortcut.com/oazo-apps/story/3790/create-piechart-component)

## Changes 👷‍♀️

- Add PieChart component.
- Add PieChart story.
- Remove faulty implementation of i18n support in Storybook.
  
## How to test 🧪

- After checking the code you can check the built storybook here: https://piechart-storybook.surge.sh/
- It looks like this:
Screen Shot 2022-04-27 at 23 02 31


blockchain/tokensMetadata.ts
```diff
@@ -15,8 +15,12 @@ export interface TokenConfig {
   tags: CoinTag[]
   color: string
   bannerIcon: string
+  bannerGif?: string
```
BackToTheCode, line 18: Are these additional props connected with this PR?

components/dumb/PieChart.tsx
```diff
@@ -0,0 +1,66 @@
...
+      angle,
+      color: items[i].color,
+    })
+    angle += 360 * ratio
+  }
+  return slices
+}
+
+export function PieChart({ items, size = 258 }: { items: PieChartItem[]; size?: number }) {
+  const strokeWidth = 34
```
BackToTheCode, line 33: Could possibly have a variable strokeWidth that scales with chart size?

BackToTheCode (approved):
Great stuff. And you fixed Storybook. Have left two thoughts

components/dumb/PieChart.tsx
BackToTheCode: Build is failing on vercel. But I can't access logs

ferni: Ah nice catch, yes it should probably scale, I'll see if I can do that.

blockchain/tokensMetadata.ts
ferni: They're not actually, just some properties that I think were missing. I attempted something using this file but then reverted, so they're kind of left overs.

sennett (approved):

PR #988, opened 2022-04-28
Automation hardhat issue fix, minor bug fixes

piekczyk:
# [Automation hardhat issue fix, minor bug fixes](https://shortcutLink)

## Changes 👷‍♀️
- item 1
  
## How to test 🧪
- step 1 ...

### DIFF
--- a/features/automation/protection/common/StopLossTriggerDataExtractor.ts
+++ b/features/automation/protection/common/StopLossTriggerDataExtractor.ts
@@ -24,9 +24,12 @@ export function extractStopLossData(data: TriggersData): StopLossTriggerData {
     // TODO: Johnnie, you shouldn't take the last one here, but rather the one that's sooner to be executed (with the highest stop loss level)
     const stopLossRecord = last(data.triggers)!
 
+    const networkId = getNetworkId()
+    const goerliOrMainnetNetwork = networkId === 5 ? 5 : 1
+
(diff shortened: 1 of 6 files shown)


features/automation/protection/common/StopLossTriggerDataExtractor.ts
```diff
@@ -24,9 +24,12 @@ export function extractStopLossData(data: TriggersData): StopLossTriggerData {
     // TODO: Johnnie, you shouldn't take the last one here, but rather the one that's sooner to be executed (with the highest stop loss level)
     const stopLossRecord = last(data.triggers)!
 
+    const networkId = getNetworkId()
+    const goerliOrMainnetNetwork = networkId === 5 ? 5 : 1
```
piotrkonowrocki, line 28: What is `5` and `1`? Can we store it as globals in this file or in some more global place? I assume it's mainnet ID and goerli ID, so maybe they can be stored in some external file for future use? If not, I suggest something like this:

```
const MAINNET_ID = 1
const GOERLI_ID = 5
// ^of from file, if possible

const networkId = getNetworkId() === GOERLI_ID ? GOERLI_ID : MAINNET_ID 
```

Thanks to that you don't have to use `goerliOrMainnetNetwork` variable which name could be misleading, because it suggest like it could have values like `'goerli' | 'mainnet'`, not their ids.

piotrkonowrocki (approved):

PR #989, opened 2022-04-28
Added ilks allowed for automation

piekczyk:
# [Added ilks allowed for automation](https://app.shortcut.com/oazo-apps/story/4195/ilk-list-for-automation-v1)

## Changes 👷‍♀️
- added ilks allowed for automation
  
## How to test 🧪
- try to use ilk that is not supported, protection tab and banners shouldn't be displayed in that case

### DIFF
--- a/features/borrow/manage/containers/ManageVaultDetails.tsx
+++ b/features/borrow/manage/containers/ManageVaultDetails.tsx
@@ -1,4 +1,4 @@
-import { ALLOWED_AUTOMATION_ILKS, getToken } from 'blockchain/tokensMetadata'
+import { getToken } from 'blockchain/tokensMetadata'
 import { DetailsSection } from 'components/DetailsSection'
 import {
   DetailsSectionContentCardWrapper,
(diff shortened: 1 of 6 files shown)


features/borrow/manage/containers/ManageVaultDetails.tsx
```diff
@@ -118,10 +118,11 @@ export function ManageVaultDetails(
   const showAfterPill = !inputAmountsEmpty && stage !== 'manageSuccess'
   const automationEnabled = useFeatureToggle('Automation')
   const automationBasicBuyAndSellEnabled = useFeatureToggle('AutomationBasicBuyAndSell')
+  const isAllowedForAutomation = ALLOWED_AUTOMATION_ILKS.includes(ilk)
 
   return (
     <Box>
-      {automationEnabled && (
+      {automationEnabled && isAllowedForAutomation && (
```
piotrkonowrocki, line 125: Suggestion: what do you thing about moving `isAllowedForAutomation` into `GetProtectionBannerControl` component? Its name is suggesting that it is controlling whenever component should be visible or not. Inside there is already a check for that (`!slData?.isStopLossEnabled && !isBannerClosed`), so it could be extended.

Rest of the components won't show up anyway unless stop loss is already turned on (`          StopLossBannerControl`) or it was just triggered (`StopLossTriggeredBannerControl`). Both won't happened since tab and banner would be off.

I'm also suggesting that because in new layout (https://github.com/OasisDEX/oasis-borrow/blob/dev/features/borrow/manage/containers/ManageVaultDetails.tsx) that stop loss banner is placed twice so if you move condition inside it, it should work in both places.

piotrkonowrocki (approved):
Everything works as intended, I just left one comment with suggestion. Feel free to follow it or resolve if you don't agree.

piotrkonowrocki (approved):

PR #1011, opened 2022-05-10
Add Warning when Stop loss is set to 3% of current collateral Ratio based on next price

piotrkonowrocki:
# [Add Warning when Stop loss is set to 3% of current collateral Ratio based on next price](https://app.shortcut.com/oazo-apps/story/4310/add-warning-when-stop-loss-is-set-to-3-of-current-collateral-ratio-based-on-next-price)

## Changes 👷‍♀️

Added box with warning when SL ratio is set in range between (NEXT_COLL_RATIO - 3) and (MAX_COLL_RATIO).
image

## How to test 🧪

Try setting up SL in vault and check if warning is visible. Ideal conditions are when next price is different than current, because in that case gap on which warning is displayed is different than three.

### DIFF
--- a/components/vault/GeneralManageLayout.tsx
+++ b/components/vault/GeneralManageLayout.tsx
@@ -28,7 +28,13 @@ export function GeneralManageLayout({
   autoTriggersData,
 }: GeneralManageLayoutProps) {
   const { t } = useTranslation()
-  const { ilkData, vault, account, priceInfo } = generalManageVault.state
+  const {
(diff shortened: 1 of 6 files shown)


johnnieskywalker (commented):
Trying on goerli but can't see the popup, let me test various edge cases, maybe it is related to my local env.
image

johnnieskywalker (approved):
image

tested on goerli view behaves as expected

image

PR #1052, opened 2022-05-23
set slippage to zero

jakubswierczek:
# [Slippage to zero](https://app.shortcut.com/oazo-apps/story/4426/bug-closing-guni-shows-dai-returned-including-slippage-which-is-false)

  
## Changes 👷‍♀️
- hardcode slippage to zero
  
## How to test 🧪
- try to close your multiply vault and look at slippage


paszkowskiDamian:
The bug appears only in guni vaults as only for those vaults we are using makers PSM where slippage is zero 😉

PR #1065, opened 2022-05-26
Gf/smart proxy redesign

George-field:
# [Smart proxy creation redesign](https://app.shortcut.com/oazo-apps/story/4221/create-smart-proxy-redesign
  
## Changes 👷‍♀️
- Added new content components for smart proxy creation
- Added new svg for displaying the current state to a user
  
## How to test 🧪
- Import the new sidebar component in OpenVaultView.tsx
- Add the below configuration

```
<SidebarSection
  title="Lorem ipsum"
  headerButton={{
   ...smartProxyHeaderConfig
  }}
  content={<SmartProxyStageOne />}
  primaryButton={{
    label: 'Primary action',
    isLoading: true,
    disabled: true,
    action: () => {
      alert('Primary action clicked')
    },
  }}
  progress={{
    text: 'Some stuff in progress',
    txHash: '',
    etherscan: '',
  }}
/>
```


components/smartProxy/SmartProxyStageTwo.tsx
```diff
@@ -0,0 +1,29 @@
...
+      </Text>
+      <Flex
+        sx={{
+          justifyContent: 'center',
+          px: '30px',
+          pt: '30px',
+          pb: '38px',
+        }}
+      >
+        <Image src="/static/img/proxy/process-setup.svg" alt="setup process" />
```
piotrkonowrocki, line 25: `src` should be in `staticFilesRuntimeUrl` method: `staticFilesRuntimeUrl('/static/img/proxy/process-setup.svg')`

components/ListWithIcon.tsx
```diff
@@ -29,14 +32,23 @@ export function ListWithIcon({
...
           >
             <Icon
               name={icon}
               size={iconSize}
-              sx={{ position: 'absolute', top: 0, left: 0, ...iconStyle }}
+              color={iconColor}
+              sx={{ position: 'absolute', top: '3px', left: 0, ...iconStyle }}
             />
-            {item}
+            <span dangerouslySetInnerHTML={{ __html: item }} />
```
piotrkonowrocki, line 51: Instead of using `dangerouslySetInnerHTML` here, you could use `Trans` tag from `next-i18next`. In the file with translation you put a value like: `Lorem <1>ipsum</1> dolor sit amet`, and then you use it using `Trans`:
```
<Trans
  i18nKey="key"
  components={{
	  1: <strong />,
  }}
/>
```
If you use it as standard `t`, `<1> will be omitted and rendered without any tag inside. What do you thing about that approach? I think it's a little more universal, because if you use that string with `strong` anywhere else, where there isn't a `dangerouslySetInnerHTML` included, it won't work.

piotrkonowrocki (commented):
I'm not going to put that as "request changes", because that's more like a questions and suggestions and I don't feel strongly enough about them that I request them.

Also, one general question. I see you updated `ListWithIcon` component as is, without `NewComponents` feature flag. That means with after first production deploy it will update the look of existing proxy page. Is that intentional?

components/ListWithIcon.tsx
George-field: That is a very good point, the reason I used it was because I tried the above and it did not work, it instead made the whole string wrap in <strong> rather than just the intended part. I may have missed something so will try again but even if we do that, any where that we do not use the 

> 

Yeah that was intentional, I did not realise this. I will create a copy of this as you make a good point.

tsconfig.json
```diff
@@ -35,7 +35,7 @@
   "include": [
     "**/*.ts",
     "**/*.tsx"
-  ],
+, "next-i18next.config.js"  ],
```
piotrkonowrocki, line 38: Do we still need that in tsconfig if you changed approach to use `Trans` component?

piotrkonowrocki (approved):

piotrkonowrocki (approved):

PR #1084, opened 2022-06-06
Stop loss form integration in open borrow flow

piekczyk:
# [Stop loss form integration in open borrow flow](https://app.shortcut.com/oazo-apps/story/4040/stop-loss-form-integration-in-borrow-vault-mvp)

## Changes 👷‍♀️
- added stop loss form to opening borrow flow
  
## How to test 🧪
- use feature toggle NewComponents and StopLossOpenFlow


George-field (commented):
looks good, not sure what we do regarding test coverage but the only comment I would have is in regards to the warnings features/borrow/open/pipes/openVault.ts here and below. Overall though, code is good :)

piotrkonowrocki (approved):
I'm approving PR as whole because it works as intended and code in general is good, but there are some code smells reported by codecov that could be easy fixed and maybe it would be worth it. For example reducing complexity, unnecessary assertions and in one place it reported that you reused name of variable that was already declared in upper scope. I don't agree with codecov about nested ternary, tho, I think often it's justified to use them.

Also, I would consider to deconstruct `priceInfo`, since you are using it in couple places and its usually as `priceInfo.someKey`.

piotrkonowrocki (approved):

piotrkonowrocki (approved):

PR #1111, opened 2022-06-10
center grid with two product cards

jakubswierczek:


paszkowskiDamian:
image

you have to solve also mobile view for this case.

```
maybe like this
columns={children.length > 2 ? [1, 2, 3] : [1, 2,children.length]}
```

PR #1113, opened 2022-06-13
Basic BS tab

piotrkonowrocki:
# Basic BS tab

Required for https://app.shortcut.com/oazo-apps/story/3869/basic-buy-form and other basic BS related stories.
  
## Changes 👷‍♀️

Created a new tab for Optimization for vaults. Currently the tab does not have any actual content. Component structure was based on Protection tab and contains some empty components, but they are going to be needed in the future.

image
  
## How to test 🧪

Go to any vault and see if Optimization tab is available.

### DIFF
--- a/components/VaultTabSwitch.tsx
+++ b/components/VaultTabSwitch.tsx
@@ -18,6 +18,7 @@ import { VaultTabTag } from './vault/VaultTabTag'
 export enum VaultViewMode {
   Overview = 'Overview',
   Protection = 'Protection',
+  Optimization = 'Optimization',
   History = 'History',
   VaultInfo = 'VaultInfo',
 }
@@ -128,6 +129,7 @@ export function VaultTabSwitch({
   const { uiChanges } = useAppContext()
   const { t } = useTranslation()
   const newComponentsEnabled = useFeatureToggle('NewComponents')
+  const basicBSEnabled = useFeatureToggle('BasicBS')
 
   useEffect(() => {
     const uiChanges$ = uiChanges.subscribe<TabChange>(TAB_CHANGE_SUBJECT)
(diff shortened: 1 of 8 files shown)


components/VaultTabSwitch.tsx
```diff
@@ -128,6 +131,7 @@ export function VaultTabSwitch({
   const { uiChanges } = useAppContext()
   const { t } = useTranslation()
   const newComponentsEnabled = useFeatureToggle('NewComponents')
+  const basicBSEnabled = useFeatureToggle('BasicBS')
```
piekczyk, line 134: we will most likely release BS and BB in the same time so this feature toggle name may be misleading

features/automation/optimization/controls/OptimizationDetailsControl.tsx
```diff
@@ -0,0 +1,16 @@
...
+function renderLayout({ vault }: { vault: Vault }) {
+  return <OptimizationDetailsLayout vault={vault} />
+}
+
+interface OptimizationDetailsControlProps {
+  vault: Vault
+}
+
+export function OptimizationDetailsControl({ vault }: OptimizationDetailsControlProps) {
+  return renderLayout({ vault })
```
piekczyk, line 15: I know that we have this pattern in stop loss (Adam's idea) but IMO render layout method add unnecessary complexity here.

piekczyk (changes requested):

components/VaultTabSwitch.tsx
piotrkonowrocki: `BasicBs` -> `Basic Buy & Sell`, at least in my mind that's what I meant, `B` in `BS` stands for `Buy`, not `Basic` 😄 What would you suggest to call it?

features/automation/optimization/controls/OptimizationDetailsControl.tsx
piotrkonowrocki: Removed.

piekczyk (approved):

PR #1135, opened 2022-06-17
Sidebar designers updates

piotrkonowrocki:
# Sidebar designers updates

All changes based on that list: https://www.notion.so/oazo/Sidebar-and-Position-page-feedback-4ad1e18bfbbd4b1eb043f734082efd5d


George-field (approved):
Looks good, only comment I have is the one about interface consistency which is an opinion

PR #1147, opened 2022-06-20
Gf/max gas fee section

George-field:
# [Max Gas Fee Component][(https://shortcutLink](https://app.shortcut.com/oazo-apps/story/4652/max-gas-fee-section))

  
## Changes 👷‍♀️
Added a new max gas fee component
modified the existing Item component to account for different props
  
## How to test 🧪
  Add the following code to a view with a sidebar

```
 <MaxGasPriceSection />
```


components/infoSection/Item.tsx
```diff
@@ -56,7 +72,7 @@ export function Item({ label, dropdownValues, value, secondaryValue, isLoading }
               <AppSpinner />
             ) : (
               <>
-                {`${value}`}
+                {value && <>{typeof value === 'object' ? value : `${value}`}</>}
```
johnnieskywalker, line 75: [minor] How about `value instanceof Object` ? 
`typeof value === 'object'` would return `true` also if value is `null`. 
However I'm not sure which is better as result of both would be very similar.

components/infoSection/Item.tsx
```diff
@@ -89,9 +105,25 @@ export function Item({ label, dropdownValues, value, secondaryValue, isLoading }
         </Box>
       </Flex>
       {open && (
-        <Grid as="ul" gap={2} sx={{ p: 0, m: 0, pl: 3, mt: 2, listStyle: 'none' }}>
-          {dropdownValues && dropdownValues.map((item) => <Item {...item} />)}
-        </Grid>
+        <>
+          {subLabel && (
+            <Text
```
piekczyk, line 110: add semantics `as="..."`

piekczyk (changes requested):

components/infoSection/Item.tsx
George-field: Ah, I was trying to find if there was something like ```React.isValidElement``` I will use that, thanks :)

piekczyk (approved):

components/infoSection/Item.tsx
johnnieskywalker: gj

PR #1148, opened 2022-06-21
Basic sell form and feature selection

piekczyk:
# [Basic sell form](https://app.shortcut.com/oazo-apps/story/3870/basic-sell-form)
# [Feature selection](https://app.shortcut.com/oazo-apps/story/4734/feature-selection-in-protection-tab)

## Changes 👷‍♀️
- added basic sell form
- added feature selection handling
  
## How to test 🧪
- use BasicBS flag (test with and without it)


johnnieskywalker (dismissed):
With BS flag on
image

switching feature is correct 
image

image

piotrkonowrocki:
Are you sure this is the right icon?

image

In previous sidebar implementations all sidebar icons were in circles. There is even similar icon for switch to borrow/multiply, but it is in a circle.

image

piotrkonowrocki:
Sam, I think this is more a quesiton to you. What should happen after uses clicks on this banner?
1. Should take user directly to stop-loss UI, if that, then @piekczyk, you need to update this button to start correct feature in protection tab.
2. Should be more general and take user to protection tab without selecting feature - if so, we need to update it's copy to be less stop-loss specific.

image

features/automation/protection/common/dropdown.ts
```diff
@@ -0,0 +1,34 @@
+import { UIChanges } from 'components/AppContext'
+import { AUTOMATION_CHANGE_FEATURE } from 'features/automation/protection/common/UITypes/AutomationFeatureChange'
+import { TFunction } from 'next-i18next'
+
+export function commonProtectionDropdownItems(uiChanges: UIChanges, t: TFunction) {
```
piotrkonowrocki, line 5: I'm not sure if I saw passing `t` as prop before, we usually use `useTranslation` hook inside of a component. Maybe update it for consistency sake?

Same for passing `uiChanges`, is there a benefit for this instead of using hook?

Rai-Stone:
> 

@piekczyk @piotrkonowrocki We should update the copy to be more generic and direct the user to the protection tab so that they can decide to use SL or BS. 

Copy can be:


### Get vault protection
Set up Stop-Loss protection or an Auto-Sell to protect your vault from if it drops below your chosen collateralization ratio.

### Protect your Vault (button)

piotrkonowrocki (approved):

PR #1158, opened 2022-06-22
Fix move events

paszkowskiDamian:
# Fix PnL for move events

Vaults https://oasis.app/27394#History and https://oasis.app/27351#Overview seem to have PnL way off due to Move events not aggregated correctly. 
  
## Changes 👷‍♀️
- Account MOVE_SRC as payback the debt
- Account MOVE_DEST as generating debt 
  
## How to test 🧪
- Checkout 27394 and 27351


johnnieskywalker (commented):
After the changes vault is not displayed in the same way. Probably it's caused by other PR that was merged to dev meanwhile.
image

In both apps same feature flags are on

image

image

Checking with new components on local environment:

No P&L is displayed for vault 27394
image

Vault 27351
same issue
image

also in old components
image

johnnieskywalker (approved):
After changes, when testing as multiply vaults :)

image


image

PR #1161, opened 2022-06-23
Update build process

paszkowskiDamian:
# [Improve release times](https://shortcutLink)
  
## Changes 👷‍♀️
- item 1
  
## How to test 🧪
- step 1 ...

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,18 +6,12 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
-
-      - name: Cache YARN dependencies
-        uses: actions/cache@v1
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
-          path: node_modules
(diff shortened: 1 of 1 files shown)


.github/workflows/aws-staging.yaml
```diff
@@ -16,82 +16,96 @@ jobs:
...
+            ${{ runner.OS }}-yarn-cache-
 
-    - name: Configure AWS credentials
-      uses: aws-actions/configure-aws-credentials@v1
-      with:
-        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-        aws-region: ${{ env.AWS_REGION }}
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile --no-progress --non-interactive
```
jeannettemcd, line 30: I believe this step will run every time, so even though you are creating cache layer, the dependencies will install anyway. You may want to add an if conditional on this step, but you will need to give Checkout step an "id" so you can reference it.

PR #1166, opened 2022-06-24
change in yields in vault header

jakubswierczek:
# [Yields change](https://app.shortcut.com/oazo-apps/story/4478/simulate-earn-header-numbers)

## Changes 👷‍♀️
* New way to get yields from API
* Change In yield (manage, open earn)

### DIFF
--- a/features/earn/yieldCalculations.ts
+++ b/features/earn/yieldCalculations.ts
@@ -1,11 +1,12 @@
 import BigNumber from 'bignumber.js'
 import { isEqual } from 'lodash'
+import moment from 'moment'
 import { combineLatest, Observable, of } from 'rxjs'
 import { catchError, distinctUntilChanged, map, switchMap } from 'rxjs/operators'
 
 import { IlkData } from '../../blockchain/ilks'
-import { one } from '../../helpers/zero'
-import { HistoricalTokenPricesApiResponse } from './makerOracleTokenPrices'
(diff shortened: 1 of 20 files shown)


features/earn/yieldCalculations.ts
```diff
@@ -1,87 +1,167 @@
...
+  const referenceDate = date || moment()
+
   return ilkData$(ilk).pipe(
     switchMap(({ ilk, token, stabilityFee, liquidationRatio }) => {
       return combineLatest(
-        makerOracleTokenPrices$(token),
+        makerOracleTokenPrices$(token, referenceDate),
+        makerOracleTokenPrices$(token, referenceDate.clone().subtract(7, 'day')),
+        makerOracleTokenPrices$(token, referenceDate.clone().subtract(30, 'day')),
+        makerOracleTokenPrices$(token, referenceDate.clone().subtract(90, 'day')),
```
sennett, line 67: What do you think of specifying what yields we want here before getting all of the token prices by default?  For example, if we only want.  Likewise for `getYieldChange$`.  As it current is it means that to get the yield change for the last day it takes 6 requests I think?

sennett (approved):

features/earn/yieldCalculations.ts
jakubswierczek: will be improved in next PR

PR #1170, opened 2022-06-27
New components headers with spacing

piotrkonowrocki:
# New components headers with spacing
  
## Changes 👷‍♀️

According to comment from Anna: all new components in details section and in sidebar should have header border with margins and footer border from side to side.

image
image


George-field (approved):
Looks good

PR #1150, opened 2022-06-21
Removed New components feature toggle

johnnieskywalker:
# [Removed new components feature toggle](https://app.shortcut.com/oazo-apps/story/4693/remove-the-new-component-feature-toggle)

<please insert a shortcut link above>
  
## Changes 👷‍♀️
  <Please add short list of changes>
- Removed new components feature toggle
- Removed some of the old unused components (ones suggested by linter)
  
## How to test 🧪
- run the app
- remove `NewComponents` and `StopLossOpenFlow` from your browser localStorage (if exists)
- new components will be visible by default, no feature flag needed.
- no old components
- no bugs
- NOTE please carefully test opening vault with SL sceanario
## Note don't merge before the release

PR #1171, opened 2022-06-27
removed remaining old components code

johnnieskywalker:
# [Cleaned remaining code from old components](https://app.shortcut.com/oazo-apps/story/4770/clean-up-old-components-code)
  
## Changes 👷‍♀️
  New components flag was already removed and merged here - https://github.com/OasisDEX/oasis-borrow/pull/1150.
 Scope of this PR is to clean up the remaining dead code and TODOs from old components.
  
## How to test 🧪
  - Check if everything is fine.
  - Let me know if there are more pieces of code that can be removed.
  - Help me clarify if `TODO` in `GeneralManageLayout.tsx` is up to date, what can be removed?

### DIFF
--- a/components/vault/GeneralManageLayout.tsx
+++ b/components/vault/GeneralManageLayout.tsx
@@ -54,7 +54,8 @@ export function GeneralManageLayout({
         defaultMode={VaultViewMode.Overview}
         heading={t(vaultHeadingKey, { ilk: vault.ilk, id: vault.id })}
         headline={headlineElement}
-        // TODO this prop to be removed when newComponentsEnabled wont be needed anymore
+        // TODO this prop to be removed when newComponentsEnabled wont be needed anymore.
+        // ^^ Not quite sure what can be removed. Isn't comment above is outdated? ~ŁW
         headerControl={<DefaultVaultHeaderControl vault={vault} ilkData={ilkData} />}
         overViewControl={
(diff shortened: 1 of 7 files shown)


components/vault/GeneralManageLayout.tsx
```diff
@@ -54,7 +54,8 @@ export function GeneralManageLayout({
         defaultMode={VaultViewMode.Overview}
         heading={t(vaultHeadingKey, { ilk: vault.ilk, id: vault.id })}
         headline={headlineElement}
-        // TODO this prop to be removed when newComponentsEnabled wont be needed anymore
+        // TODO this prop to be removed when newComponentsEnabled wont be needed anymore.
+        // ^^ Not quite sure what can be removed. Isn't comment above is outdated? ~ŁW
```
George-field, line 58: Yeah this looks like an outdated comment, may be best to just remove it to avoid future confusion.

George-field (changes requested):
All good, just think we need to remove the comment you mentioned as it is outdated and will cause confusion :)

components/vault/GeneralManageLayout.tsx
johnnieskywalker: Ok let's remove it

George-field (approved):
All good :)

PR #1177, opened 2022-06-28
Update vulnerable packages 

piotrkonowrocki:
# Update vulnerable packages

There are two packages left with critical errors.

### xmlhttprequest-ssl
Used in: `@myetherwallet/mewconnect-connector`.
Chain of dependency: `@myetherwallet/mewconnect-connector` > `@myetherwallet/mewconnect-web-client` > `socket.io-client` > `engine.io-client` > `xmlhttprequest-ssl`

This package is deprecated, so it may takes some time to fix it.

### json-schema
Used in: `snowflake-sdk`, `node-sass`, `@oasisdex/web3-context`, `@oasisdex/connectors`, `web3`, `@oasisdex/connectors`, `@myetherwallet/mewconnect-connector`.
This package is used seven dependencies, but the problem is that it is a deep dependency of dependency of dependency (...), that always ends in the same place. This package in problematic version is used in `request` which is deprecated and even newest version causes this error. Some of the dependencies are our internal commons from `@oasisdex` and maybe they don't need `request` at all, but that doesn't solve others. Needs deep investigation if they can be updated.
  
## Changes 👷‍♀️

- linked `ejs` to resolution in version `^3.1.7`

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,6 +145,7 @@
     "zod": "^1.5.0"
   },
   "resolutions": {
+    "ejs": "^3.1.7",
     "minimist": "^1.2.6",
     "mixin-deep": "^1.3.2",
     "mkdirp": "^0.5.5",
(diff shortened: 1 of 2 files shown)


johnnieskywalker:
MEW has new connector, so fix will be to use new dependency https://app.shortcut.com/oazo-apps/story/3984/use-new-mew-connector

johnnieskywalker (approved):

PR #1185, opened 2022-06-30
Basic sell set min price validation

piekczyk:
# [Basic sell set min price validation](https://app.shortcut.com/oazo-apps/story/4989/validation-basic-sell-set-min-sell-price)

## Changes 👷‍♀️
- added validation with warning message when user tries to set min sell price equal to zero or uses no threshold
  - fixed issue with eth market price
## How to test 🧪
- using BasicBS flag try to add basic sell trigger with and without setting min sell price (do it in each possible case - when adding fresh trigger and updating existing one)


johnnieskywalker:
image

image

when form is in default state and user sets no threshold button is blocked when min price is 0
image


I'm not 100% sure if this is correct scenario, price is set to 0, button is blocked
image
but then user clicks no threshold and button gets unlocked
image
 On existing trigger it's not allowed to set up again trigger with the same values with and without threshold
 image
image

johnnieskywalker (dismissed):
image
without threshold there is warning but button is enabled
image

johnnieskywalker (approved):

PR #1190, opened 2022-07-01
changed max gas fee to gwei instrad of percentage

George-field:
# [Max Gas Fee UI Updates](https://app.shortcut.com/oazo-apps/story/4998/max-gas-fee-handling-on-the-front-end-side)

  
## Changes 👷‍♀️
 - Updated the UI for max gas fee to now be represented by a gwei rather than a %


johnnieskywalker (changes requested):
Changes is ok, as scope of story is just UI.
However state initialization in `useBasicSellStateInitializator.ts` is missing.
Moreover `useBasicSellStateInitializator` is used both for basic sell and buy.
I'm not quite sure if it will remain that way so we could move it to common, or it will be split into 2 different initializators.
IMHO state should be also initialized in scope of that story so also this small refactor could be done. Do you agree @piekczyk ?

piekczyk:
> 
when it comes to UI change its enough, logic part including initialization and handling is in the scope of other story if I recall correctly. 

`useBasicSellStateInitializator` we should just change file name (I guess we can do it at the very end to avoid conflicts), this file contains initialization both for basic buy and basic sell using given `type` so there is no need for duplication

johnnieskywalker (dismissed):

PR #1228, opened 2022-07-08
Action for staging rollback

jakubswierczek:


cristidas (approved):

.github/workflows/aws-staging-rollback.yaml
```diff
@@ -0,0 +1,64 @@
+name: Rollback Staging
+
+on:
+  workflow_dispatch:
+    branches: [ dev ]
+    inputs:
+      commit_hash:
+        description: 'Enter the commit hash you want to roll back to.'
```
sennett, line 8: Is there any more info we can give here?  I think it has to be from a specific branch?  One that has an image built for it?  Could also specify the short commit hash?  Update the rollback-prod one too?

sennett (dismissed):

paszkowskiDamian (approved):

cristidas (approved):

PR #1245, opened 2022-07-12
Gf/notificaitions icon

George-field:
# [Notifications Icon](https://app.shortcut.com/oazo-apps/story/5096/notification-center-ui-component-create-notification-centre-icon)

  
## Changes 👷‍♀️
- Updated header item spacing to match new design
- Added a new notification icon button component that displays a count and triggers the opening and closing of notifications
  
## How to test 🧪
Turn on feature toggle (notifications) and refresh the page. Make sure your wallet is connected and you will see the icon in the top right.


helpers/useFeatureToggle.ts
```diff
@@ -22,6 +23,7 @@ const configuredFeatures: Record<Feature, boolean> = {
   StopLossWrite: true,
   BasicBS: false,
   ReadOnlyBasicBS: false,
+  notifications: false,
```
piekczyk, line 26: dunno why we are using capital letter but I guess we should stick to it, so `Notifications`

piekczyk (changes requested):

helpers/useFeatureToggle.ts
George-field: Ahh, good point

piekczyk (approved):

PR #1286, opened 2022-07-19
Gf/notification prefrences

George-field:
# [Notifications Preferences](https://app.shortcut.com/oazo-apps/story/4887/manage-notifications-ui-component-notifications-on-off-switch)

  
## Changes 👷‍♀️
- Added notifications preferences cards & section
  
## How to test 🧪
- Enable notifications by turning on the feature toggle in local storage. Click the notification Icon in the top right, then select the settings icon in the top right of the notifications modal.


components/notifications/NotificationPrefrenceCard.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react'
+import { Box, Card, Flex, Text } from 'theme-ui'
+import { Toggle } from 'ui'
+
+interface NotificationPrefrenceCardPtops {
```
johnnieskywalker, line 5: [minor][ a typo, `Props` is expected :D

johnnieskywalker (changes requested):

components/notifications/NotificationPrefrenceCard.tsx
George-field: ah, dam it, thank you

ui/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,65 @@
...
+        }}
+        type="checkbox"
+        onChange={() => setChecked(!checked)}
+      />
+      <Box
+        as="span"
+        sx={{
+          '&:before': {
+            position: 'absolute',
+            // TODO: Possibly create a function to do this so that is more readable
```
piekczyk, line 36: I agree with TODO, if there is no other way to define these svgs we should definitely extract it to constant / function which returns it

components/notifications/NotificationsCenter.tsx
```diff
@@ -52,7 +54,12 @@ export function NotificationsCenter({ isOpen }: { isOpen: boolean }) {
       <NotificationsCenterContent>
         <Grid sx={{ px: 2, mt: 3 }}>
           {showPrefrencesTab ? (
-            <p>Prefrences</p>
+            <Box>
```
piekczyk, line 57: maybe it is worth to extract it as for example `NotificationPreferencesWrapper` even though it is simple box for now

piekczyk (changes requested):

piekczyk (approved):

johnnieskywalker (approved):

PR #1294, opened 2022-07-20
Cache & batch Infura calls on Server

BackToTheCode:
# [Off-chain cache for Infura calls](https://app.shortcut.com/oazo-apps/story/5172/off-chain-cache-for-infura-calls)

Reduces infura calls for concurrent sessions. Current TTL set at 15 seconds but can be increased to further reduce infura call volume at the cost of slightly stale data in app.
  
## Changes 👷‍♀️

- Add new custom JsonRpcProvider called `JsonRpcCacheProvider` that handles the batching of calls and sends them across to our new api endpoint
- Add new endpoint that handles unbatching and caching calls and rebatching cache misses before making a batched call to infura. Reconstitutes all results and passes back to client
- Put caching functionality behind a feature toggle for testing on staging
- Falls back to previous custom batching provider `JsonRpcBatchProvider` when flag is disabled and when using hardhat in staging or on vercel preview deployment
- Standardises all infura eth_calls to use the MCD_Spot contract address in the from field. This is required for OSMS calls. But, it also means all read_only calls are consistent between users and the cache key currently includes the from field when generating a digest.
  
## How to test 🧪
Find `pages/api/infuraCallsCache.tsx` and update line 24: `const cache = new BatchCache()` to this ` const cache = new BatchCache({ debug: true })`. Cache TTL can also be increased here to aid test
Toggle the feature flag to true for `BatchCache in your browser console.

- Navigate to various marketing pages - check network tab and see infura calls - at the same time open 2 other tabs in other browsers and request the same page in those browsers. You should see cache hits appearing in the console assuming requests from different tabs are arriving within the 15s default window.
- Open a multiply vault & earn vault - confirm all positions were opened without issue

---- ---- ----
**Testing with multiple browsers**

**1 Concurrent Sessions**
Screenshot 2022-07-22 at 13 09 19

**2 Concurrent Sessions**
Screenshot 2022-07-22 at 13 09 32

**2 + Additional 3 Sessions within 15 seconds TTL window**
Screenshot 2022-07-22 at 13 09 45


blockchain/calls/callsHelpers.ts
```diff
@@ -37,10 +37,8 @@ export function callAbstractContext<D, R, CC extends Context>(
         args,
         context,
       )(...prepareArgs(args, context)).call(
-        context.status === 'connected'
-          ? { from: (context as any).account }
-          : // spot neccessary to read osms in readonly
-            { from: context.mcdSpot.address },
+        // spot neccessary to read osms in readonly
+        { from: context.mcdSpot.address },
```
paszkowskiDamian, line 41: Did you test sending tx with such setup?

helpers/api/BatchManager.test.ts
```diff
@@ -0,0 +1,247 @@
...
+  })
+
+  it('should batch call into single request', async () => {
+    const mockBatch = createMockBatch()
+    const mockFetchJson = sinon.stub().resolves([]) as any
+
+    const batchManager = createBatchManager(mockFetchJson)
+    await batchManager.batchCall(mockBatch)
+
+    expect(mockFetchJson).to.have.been.calledWith(mockConnection, JSON.stringify(mockBatch))
```
sennett, line 27: Could assert that `mockFetchJson` was only called once here.

helpers/api/BatchManager.test.ts
```diff
@@ -0,0 +1,247 @@
...
+    const batchManager = createBatchManager(mockFetchJson)
+
+    // First call
+    await batchManager.batchCall(mockBatch)
+    mockFetchJson.reset()
+
+    // Second call
+    await batchManager.batchCall(mockBatch)
+
+    expect(mockFetchJson).to.not.have.been.called
```
sennett, line 47:
```suggestion
    expect(mockFetchJson).to.have.been.called.once
```
and remove the `reset` above?

helpers/api/BatchManager.test.ts
```diff
@@ -0,0 +1,247 @@
+import { expect } from 'chai'
+import sinon from 'sinon'
+
+import { BatchCache } from './BatchCache'
+import { BatchManager, Request } from './BatchManager'
+
+const mockConnection = 'https://www.goingnowhere.com'
+
+describe('BatchManager', () => {
```
sennett, line 9: Awesome stuff.  Could we have a bunch of testing around the ordering of the results from `BatchManager`?  Both 100% from cache, 0% from cache, and interleaving the results?

sennett (changes requested):

helpers/api/skipCache.ts
```diff
@@ -0,0 +1,20 @@
+import { networksByName } from 'blockchain/config'
+import { useFeatureToggle } from 'helpers/useFeatureToggle'
+
+function isLocalhost(url: string) {
+  return url.includes('localhost') || url.includes('127.0.0.1')
+}
+
+function isUsingHardhat(chainId: string) {
+  return (networksByName['hardhat'].id = chainId)
```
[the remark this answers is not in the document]
BackToTheCode: Errr.. ya 🤦‍♂️. Thank you for the spot

sennett (approved):

PR #1295, opened 2022-07-20
As/gas estimation refactor

adamskrodzki:
Refactoring of gasEstimation, idea is that in our app only one transaction get estimated at the time, and all estimations look the same really, so instead of repeating estimation everywhere context is provided

uiChanges.publish is used to push data (txDefinition and payload) that are supposed to be estimated

And provider in return estimate transaction and provides context


johnnieskywalker (dismissed):
It's more optimal, next step would be to refactor usages of old method

johnnieskywalker (approved):

components/vault/VaultChangesInformation.tsx
```diff
@@ -112,10 +110,36 @@ export function getEstimatedGasFeeText(gasEstimation?: HasGasEstimation, withBra
...
+  }
+  const status: GasEstimationStatus = gasEstimation.isCompleted
+    ? gasEstimation.isSuccessful
+      ? GasEstimationStatus.calculated
+      : GasEstimationStatus.error
+    : GasEstimationStatus.calculating
+
+  switch (status) {
+    case GasEstimationStatus.calculating:
+      const textPending = 'Pending...'
```
piotrkonowrocki, line 133: Translations?

features/gasEstimation/GasEstimation.tsx
```diff
@@ -0,0 +1,9 @@
+import { useGasEstimationContext } from 'components/GasEstimationContextProvider'
+import { getEstimatedGasFeeText } from 'components/vault/VaultChangesInformation'
+import React from 'react'
+
+export function GasEstimation() {
```
adamskrodzki, line 5: shouldn't that be component instead of feature ?

components/vault/VaultChangesInformation.tsx
```diff
@@ -112,10 +110,36 @@ export function getEstimatedGasFeeText(gasEstimation?: HasGasEstimation, withBra
...
+
+      return (
+        <Text sx={{ color: 'neutral80' }}>{withBrackets ? `(${textPending})` : textPending}</Text>
```
piotrkonowrocki, line 136: Should be rendered as `span` or at least `p`.

piotrkonowrocki (approved):

PR #1308, opened 2022-07-22
Gf/notifications email prefrences

George-field:
# [Notification Preferences](https://app.shortcut.com/oazo-apps/story/4891/email-notifications-ui-component)

  
## Changes 👷‍♀️
  - Added notifications email preferences section
  - Added email validation function
  - Added notifications email preferences cards for changing email and having a valid email
  
## How to test 🧪
Go to notifications preferences by clicking on the notifications icon, clicking the settings icon and then toggle the email notifications toggle to true


components/notifications/NotificationsChangeEmailButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+      >
+        {currentEmail}
+      </Text>
+
+      <Button
+        sx={{
+          color: '#575CFE',
+          fontSize: '14px',
+          background: 'transparent',
+          marginLeft: 'auto !important',
```
piotrkonowrocki, line 38: Is that necessary? `!important` should be last resorts, because from that point it could only go downhill. What is default value of that margin and is it possible to overwrite it without using importants?

components/notifications/NotificationsEmailPrefrences.tsx
```diff
@@ -0,0 +1,93 @@
...
+              borderRadius: '8px',
+            }}
+          >
+            <Text
+              sx={{
+                fontSize: '12px',
+                fontWeight: 600,
+              }}
+            >
+              Set up your email
```
piotrkonowrocki, line 43: Shouldn't that be in translations?

components/notifications/NotificationsSetupSuccess.tsx
```diff
@@ -0,0 +1,51 @@
...
+    >
+      <Flex
+        sx={{
+          justifyContent: 'center',
+          width: '50px',
+          height: '50px',
+          alignItems: 'center',
+        }}
+      >
+        <svg
```
piotrkonowrocki, line 26: Consider moving to global icons?

helpers/formValidation.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Validate an email address
+ * @param email
+ * @returns
+ */
+export function validateEmail(email: string) {
+  return email.match(
+    /^(([^<>()[\]\\.,;:\s@]+(\.[^<>()[\]\\.,;:\s@]+)*)|(.+))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/,
```
piotrkonowrocki, line 8: I'm wondering about that regex, have you used it somewhere before? I've been using different and it's much shorter: `^[\w-\.]+@([\w-]+\.)+[\w-]{2,4}$`. What do you think, is it missing something?

piotrkonowrocki (changes requested):
I have some issues with css/sx code, but there are recurring errors and I don't want to mark every one of them as separated comment.

Please update css to use more of our theme. There are a lot of colors or border radius values that are hardcoded, but should be taken from theme file. I also notices some of font sizes that are available in theme, like 8 or 16px, that are also hardcoded.

About some of the animations, what exactly is transitioning in them? Not specifying css selectors that suppose to transition can sometimes lead to unexpected behaviors, I suggest listing what exactly is animating instead of using `all` or providing just a duration.

Also, that's only suggestion, not request change, but I suggest using shorthands in `sx` objects, `marginLeft` -> `ml`, `paddingLeft` -> `pl` etc.

components/notifications/NotificationsChangeEmailButton.tsx
George-field: It is being overridden by an !important somewhere else in the code :(, I believe it is a default style that is being applied by the UI library.

components/notifications/NotificationsEmailPrefrences.tsx
George-field: yes, I missed that one

helpers/formValidation.ts
George-field: I'll go with your one, I just found that on stackoverflow if I am honest

components/notifications/NotificationsChangeEmailButton.tsx
piotrkonowrocki: That's exactly what I meant by saying that it can only go downhill from there. Someone already did that mistake earlier and it's forcing you to using `important` because of that. In that case I understand that it can't be done in any other way.

components/notifications/NotificationsSetupSuccess.tsx
```diff
@@ -0,0 +1,39 @@
...
+import React from 'react'
+import { Flex, Text } from 'theme-ui'
+
+export function NotificationsSetupSuccess() {
+  const { t } = useTranslation()
+
+  return (
+    <Flex
+      sx={{
+        p: '16px',
```
piekczyk, line 12: 3

components/notifications/NotificationsSetupSuccess.tsx
```diff
@@ -0,0 +1,39 @@
+import { Icon } from '@makerdao/dai-ui-icons'
+import { useTranslation } from 'next-i18next'
+import React from 'react'
+import { Flex, Text } from 'theme-ui'
+
+export function NotificationsSetupSuccess() {
```
piekczyk, line 6: this component name is too general since it has hardcoded email-related copy

components/notifications/NotificationsEmailPrefrences.tsx
```diff
@@ -0,0 +1,89 @@
...
+      {enableEmailPrefrences && (
+        <Box
+          sx={{
+            px: '5px',
+          }}
+        >
+          <Box
+            sx={{
+              background: 'neutral30',
+              p: '16px',
```
piekczyk, line 29: 3

piekczyk (changes requested):

piotrkonowrocki (approved):

PR #1312, opened 2022-07-22
[sc-4824] Creating proxy with xstate in open-aave

jakubswierczek:
# [Opening AAVE Position with ETH](https://app.shortcut.com/oazo-apps/story/4824/open-position-form-with-eth)
## Changes 👷‍♀️
  The new approach to state handling
- usage of `xstate`
- refactor the proxy creation
- new context (nested context used only in OpenAave) - our context (AppContext) is too big. 
  
## How to test 🧪
- go to `/earn/open/aave-steth`
- try to create a proxy with the new view.


features/proxyNew/state/proxyStateMachine.types.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export const PROXY_STAGES = [
+  'proxyIdle',
+  'proxyWaitingForApproval',
+  'proxyInProgress',
+  'proxyFailure',
+  'proxySuccess',
+] as const
+
+export type ProxyStages = typeof PROXY_STAGES[number]
```
sennett, line 59: Could merge these two and remove unnecessary export.

```suggestion
export const PROXY_STAGES = [
  'proxyIdle',
  'proxyWaitingForApproval',
  'proxyInProgress',
  'proxyFailure',
  'proxySuccess',
] as const

type ProxyStateKeys = 'proxyIdle' |
  'proxyWaitingForApproval' |
  'proxyInProgress' |
  'proxyFailure' |
  'proxySuccess'
```

Also I'm not sure of the purpose of these types.  I can do this:

image

and it compiles fine.  What is this type for?

features/proxyNew/state/proxyStateMachine.types.ts
```diff
@@ -0,0 +1,76 @@
...
+  readonly txHelper: TxHelpers
+  readonly proxyAddress$: Observable<string | undefined>
+  readonly getGasEstimation$: (estimatedGasCost: number) => Observable<HasGasEstimation>
+  readonly safeConfirmations: number
+  gasData?: HasGasEstimation
+  txHash?: string
+  txError?: string
+  proxyConfirmations?: number
+  proxyAddress?: string
+}
```
sennett, line 18: What do you think of grouping this by dependencies and machine-state?

```suggestion
export interface ProxyContext {
  deps: {
    readonly txHelper: TxHelpers
    readonly proxyAddress$: Observable<string | undefined>
    readonly getGasEstimation$: (estimatedGasCost: number) => Observable<HasGasEstimation>
    readonly safeConfirmations: number
  },
  inMachine: {
    gasData?: HasGasEstimation
    txHash?: string
    txError?: string
    proxyConfirmations?: number
    proxyAddress?: string
  }
}
```

`inMachine` is a pretty crappy name but what do you think of the idea?

features/earn/aave/open/state/openAaveStateMachine.ts
```diff
@@ -0,0 +1,175 @@
...
+            assign({
+              totalSteps: (context) => (context.proxyAddress ? 2 : 3),
+            }),
+            assign({
+              currentStep: (_) => 1,
+            }),
+            assign({
+              canGoToNext: (context) =>
+                context.proxyAddress === undefined || context.amount?.gt(zero),
+            }),
```
sennett, line 39: Is it possible to group all these into one `assign`?

features/earn/aave/AaveContext.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const proxyForAccount$: Observable<string | undefined> = contextForAddress$.pipe(
+    switchMap(({ account }) => proxyAddress$(account)),
+  )
+
+  const proxyStateMachine$ = curry(
+    memoize(() =>
+      getOpenProxyStateMachine$(contextForAddress$, txHelpers$, proxyForAccount$, gasEstimation$),
+    ),
+  )()
```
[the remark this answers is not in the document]
jakubswierczek: yes, you are right

features/proxyNew/state/proxyStateMachine.services.ts
```diff
@@ -0,0 +1,75 @@
...
+    map((gas) => ({ type: 'GAS_COST_ESTIMATION', gasData: gas })),
+  )
+}
+
+export const services = {
+  createProxy,
+  estimateGas,
+}
+
+export const getNameOfService = (name: Extract<keyof typeof services, string>): string => name
```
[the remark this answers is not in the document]
jakubswierczek: I forgot to use it. See the next comment for an explanation.

features/proxyNew/state/proxyStateMachine.ts
```diff
@@ -0,0 +1,123 @@
...
+            RETRY: {
+              target: 'proxyInProgress',
+            },
+          },
+        },
+      },
+    },
+    {
+      services: {
+        ...services,
```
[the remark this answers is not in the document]
jakubswierczek: Yes, you are right. The invocation isn’t typesafe. That’s why I’ve created `getNameOfService` function. 
With that we are sure that we getting the right service:
```
src: getNameOfService('estimateGas’)
```
You can easily navigate to `estimateGas` with that function. 

I prefer having invocations in the `service` node because it is easy to mock or change with `.withConfig(…)` method for a  machine:
https://xstate.js.org/docs/guides/machines.html#extending-machines
Great to use in tests.

features/proxyNew/state/proxyStateMachine.types.ts
jakubswierczek: you can, it is an issue with types in the library I think. 
types are useful when you are checking the state of the machine:
```
state.matches('proxyIdle'):
```

FYI: unnecessary export removed.

features/proxyNew/state/proxyStateMachine.types.ts
jakubswierczek: great idea!

features/earn/aave/open/state/openAaveStateMachine.ts
jakubswierczek: yes

features/earn/aave/open/state/openAaveStateMachine.ts
```diff
@@ -0,0 +1,175 @@
...
+            assign({
+              totalSteps: (context) => (context.proxyAddress ? 2 : 3),
+            }),
+            assign({
+              currentStep: (_) => 1,
+            }),
+            assign({
+              canGoToNext: (context) =>
+                context.proxyAddress === undefined || context.amount?.gt(zero),
```
sennett, line 38: Assuming that `context.amount` is set we can always go to next.  If the proxy address is set, then we go to the confirm screen. If not, then proxy setup sub state-machine.  We could use guarded transitions here I think (https://xstate.js.org/docs/guides/guards.html#guarded-transitions)?  Or we could send a next function to the view transition/next state function to the view (similar `progress`/`regress` in current app).

features/proxyNew/state/proxyStateMachine.types.ts
jakubswierczek: I put the dependencies in a separated node, but it will be better to keep the state in root because we can't simply `assign` nested object:
	```javascript
	assign((context, event) => ({  
	  machineState: {  
	    proxyAddress: event.proxyAddress,  
	    totalSteps: event.proxyAddress ? 2 : 3,  
	  },  
	})),
	```
The code will replace `machineState` instead of merging with the initial state

features/earn/aave/open/state/openAaveStateMachine.ts
jakubswierczek: yes.
```
CREATE_PROXY: {
               target: 'proxyCreating',
               cond: (context, _) => context.proxyAddress === undefined,
             },
``` 
That transition is guarded and in view we can do:
```
 state.can('CREATE_PROXY')
```

features/proxyNew/state/proxyStateMachine.services.ts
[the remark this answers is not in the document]
paszkowskiDamian: isn't `keyof typeof services` returning all strings any way so this Extract looks like redundant type 🤔

features/earn/aave/AaveContext.ts
```diff
@@ -0,0 +1,45 @@
+import { curry } from 'ramda'
+import { Observable } from 'rxjs'
+import { distinctUntilKeyChanged, switchMap } from 'rxjs/operators'
+
+import { AppContext } from '../../../components/AppContext'
+import { getGasEstimation$ } from '../../proxyNew/pipelines/getGasEstimation'
+import { getOpenProxyStateMachine$ } from '../../proxyNew/pipelines/getOpenProxyStateMachine'
+import { getOpenAaveStateMachine } from './pipelines/getOpenAaveStateMachine'
+
+export function setupAaveContext({
```
BackToTheCode, line 10: Love this

features/proxyNew/state/proxyStateMachine.ts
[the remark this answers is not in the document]
BackToTheCode: We should be able to make everything typesafe when declaring services, actions and guards. See -> https://xstate.js.org/docs/guides/typescript.html#typegen

Should also address all the type issues mentioned in other comments too.

You can see an example in this branch -> `jt/multi-step-form-poc`

Using VSCode we can generate a typegen file that'll make services, actions, and guards typesafe. Not sure what the equivalent extension is for webstorm.

We then import that file. Here's an example when creating a state-machine

```
const borrowPositionStateMachine = createMachine(
  {
    schema: {
      context: {} as BorrowPositionMachineContext,
      events: {} as BorrowPositionMachineEvent,
      services: {} as {
        proxyService: { data: { type: 'PROXY_CREATED'; info: ProxyInfo } }
        allowanceService: { data: { type: 'ALLOWANCE_SET'; info: AllowanceInfo } }
        openPosition: { data: { type: 'POSITION_CREATED'; info: PositionInfo } }
      },
    },
    tsTypes: {} as import('./index.typegen').Typegen0, // HERE
```

features/earn/aave/open/sidebars/GetSidebarTexts.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const canCreateProxy = state.can('CREATE_PROXY')
+
+  const baseSettings = {
+    steps,
+    isLoading,
+    url,
+    disabled: !state.context.canGoToNext,
+  }
+
+  switch (true) {
```
BackToTheCode, line 33: Could we replace this with 

```
switch (state.value):
  case states.EDITING:
```

features/earn/aave/open/sidebars/SidebarOpenAaveVault.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const isFailed = state.matches('txFailure')
+
+  const texts = GetSidebarTexts(state, t, send)
+
+  const sidebarSectionProps: SidebarSectionProps = {
+    title: t('open-earn.aave.vault-form.title'),
+    content: (
+      <Grid gap={3}>
+        {state.matches('editing') && (
```
BackToTheCode, line 32: For all the state matches could we use an object/enum

features/proxyNew/state/proxyStateMachine.ts
```diff
@@ -0,0 +1,109 @@
...
+        },
+        proxyInProgress: {
+          invoke: {
+            src: getNameOfService('createProxy'),
+          },
+          on: {
+            SUCCESS: {
+              target: 'proxySuccess',
+              actions: [
+                assign((_, event) => ({ proxyAddress: event.proxyAddress })),
```
BackToTheCode, line 55: Guess we don't need?

features/proxyNew/state/proxyStateMachine.ts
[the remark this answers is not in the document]
BackToTheCode: This may be solved by comment above?

BackToTheCode (commented):
Boom. This is awesome

features/earn/aave/open/state/openAaveStateMachine.ts
```diff
@@ -0,0 +1,149 @@
...
+                })),
+              ],
+            },
+            CREATE_PROXY: {
+              target: 'proxyCreating',
+              cond: (context, _) => context.proxyAddress === undefined,
+            },
+            CONFIRM_DEPOSIT: {
+              target: 'reviewing',
+              cond: (context, _) => context.proxyAddress !== undefined,
```
BackToTheCode, line 71: Could we use multiple guards to this event so

https://xstate.js.org/docs/guides/guards.html#multiple-guards

```suggestion
              cond: (context, _) => context.proxyAddress !== undefined && context.amount.gt(0),
```

features/earn/aave/open/sidebars/GetSidebarTexts.tsx
```diff
@@ -0,0 +1,58 @@
...
+    : undefined
+
+  const canCreateProxy = state.can('CREATE_PROXY')
+
+  const baseSettings = {
+    steps,
+    isLoading,
+    url,
+    disabled: !state.context.canGoToNext,
+  }
```
BackToTheCode, line 31: I found `canGoToNext` slightly confusing just because we have transition guards. Appreciate this is for use in disabling button.

What do you think about following maybe?


```suggestion
  const canCreateProxy = state.can(states.CREATE_PROXY)
  const canConfirmDeposit = state.can(states.CONFIRM_DEPOSIT)
  const canProceed = canCreateProxy || canConfirmDeposit

  const baseSettings = {
    steps,
    isLoading,
    url,
    disabled: !canProceed,
  }
```

BackToTheCode (approved):

PR #1326, opened 2022-07-29
Auto BS error when CM enabled

piekczyk:
# [Auto BS error when CM enabled](https://app.shortcut.com/oazo-apps/story/5240/ab-as-form-validation-cannot-add-auto-buy-or-auto-sell-if-the-user-has-constant-multiple)

## Changes 👷‍♀️

- added error to auto buy and sell forms when user tries to add trigger when CM is enabled
  
## How to test 🧪

- enable CM and then try to add auto sell or buy trigger, you should get error (not testable due to lack data from cache)

### DIFF
--- a/components/vault/VaultErrors.tsx
+++ b/components/vault/VaultErrors.tsx
@@ -20,6 +20,7 @@ interface VaultErrorsProps {
   maxGenerateAmount?: BigNumber
   ilkData: IlkData
   maxWithdrawAmount?: BigNumber
+  autoType?: 'Auto-Buy' | 'Auto-Sell'
 }
 
 export function VaultErrors({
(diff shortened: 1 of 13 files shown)


piotrkonowrocki (approved):

components/vault/VaultErrors.tsx
```diff
@@ -20,13 +20,15 @@ interface VaultErrorsProps {
   maxGenerateAmount?: BigNumber
   ilkData: IlkData
   maxWithdrawAmount?: BigNumber
+  autoType?: 'Auto-Buy' | 'Auto-Sell'
```
johnnieskywalker, line 23: I kinda don't like the fact that it's only used by automation while VaultErros is used in multiple places. But as it's optional it seems like proper way of handling it.

PR #1327, opened 2022-07-29
Constant multiple changes information data

piotrkonowrocki:
# [Constant multiple changes information data](https://app.shortcut.com/oazo-apps/story/3876/constant-multiple-changes-information)
  
## Changes 👷‍♀️

- Connected changes information to `constantMultipleState`,
- updated `getBasicBSVaultChange` method to accept constant multiple values.

image

To do in the future, in next PR:
- Decide if `slippage` needs to be removed.
  
## How to test 🧪

- Check if changes information are updating in constant multiple form.

### DIFF
--- a/features/automation/basicBuySell/InfoSections/ConstantMultipleInfoSection.tsx
+++ b/features/automation/basicBuySell/InfoSections/ConstantMultipleInfoSection.tsx
@@ -18,6 +18,8 @@ interface ConstantMultipleInfoSectionProps {
   collateralToBeSold: BigNumber
   minPriceToSell?: BigNumber
   addTriggerGasEstimationUsd?: BigNumber
+  estimatedOasisFee: BigNumber[]
+  estimatedGasPriceOnTrigger?: BigNumber
 }
 
 export function ConstantMultipleInfoSection({
@@ -34,6 +36,8 @@ export function ConstantMultipleInfoSection({
   collateralToBeSold,
   minPriceToSell,
   addTriggerGasEstimationUsd,
+  estimatedOasisFee,
+  estimatedGasPriceOnTrigger,
 }: ConstantMultipleInfoSectionProps) {
   const { t } = useTranslation()
 
@@ -55,8 +59,29 @@ export function ConstantMultipleInfoSection({
         },
         {
           label: t('constant-multiple.vault-changes.cost-per-adjustment'),
-          // TODO: PK calculate this value
-          value: '$0',
+          value:
+            estimatedGasPriceOnTrigger &&
+            estimatedOasisFee
+              .map((feeItem) => {
+                return `$${formatAmount(feeItem.plus(estimatedGasPriceOnTrigger), 'USD')}`
+              })
+              .join(' - '),
+          isLoading: estimatedGasPriceOnTrigger === undefined,
+          dropdownValues: [
+            {
+              label: t('constant-multiple.vault-changes.estimated-oasis-fee'),
+              value: estimatedOasisFee
+                .map((feeItem) => {
+                  return `$${formatAmount(feeItem, 'USD')}`
+                })
+                .join(' - '),
+            },
+            {
+              label: t('constant-multiple.vault-changes.estimated-max-gas-fee'),
+              value:
+                estimatedGasPriceOnTrigger && `$${formatAmount(estimatedGasPriceOnTrigger, 'USD')}`,
+            },
+          ],
         },
         {
           label: t('auto-sell.setup-transaction-cost'),
--- a/features/automation/basicBuySell/InfoSections/ConstantMultipleInfoSection.tsx
+++ b/features/automation/basicBuySell/InfoSections/ConstantMultipleInfoSection.tsx
@@ -60,7 +60,8 @@ export function ConstantMultipleInfoSection({
         },
         {
           label: t('auto-sell.setup-transaction-cost'),
-          value: addTriggerGasEstimationUsd && `$${formatAmount(addTriggerGasEstimationUsd, 'USD')}`,
+          value:
+            addTriggerGasEstimationUsd && `$${formatAmount(addTriggerGasEstimationUsd, 'USD')}`,
           isLoading: addTriggerGasEstimationUsd === undefined,
         },
         {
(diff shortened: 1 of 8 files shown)


features/automation/optimization/sidebars/SidebarSetupConstantMultiple.tsx
```diff
@@ -304,30 +312,44 @@ function ConstantMultipleInfoSectionControl({
...
   return (
     <ConstantMultipleInfoSection
       token={token}
-      targetColRatio={targetColRatio}
-      multiplier={multiplier}
+      targetColRatio={constantMultipleState.targetCollRatio}
+      multiplier={constantMultipleState.multiplier}
       slippage={slippage}
-      triggerColRatioToBuy={triggerColRatioToBuy}
+      triggerColRatioToBuy={constantMultipleState.buyExecutionCollRatio}
```
piekczyk, line 334: maybe it is worth to rename props to names from constantMultipleState? So in that case `buyExecutionCollRatio`? In that we will avoid calling same thing by different names.

features/automation/optimization/sidebars/SidebarSetupConstantMultiple.tsx
```diff
@@ -304,30 +312,44 @@ function ConstantMultipleInfoSectionControl({
...
-  const targetColRatio = new BigNumber(200)
-  const multiplier = 2
+  // TODO: PK where do I get slippage?
   const slippage = new BigNumber(0.5)
-  const triggerColRatioToBuy = new BigNumber(200)
-  const triggerColRatioToSell = new BigNumber(300)
-  const maxPriceToBuy = new BigNumber(1600)
-  const minPriceToSell = new BigNumber(1200)
+  const feeDiff = estimatedBuyFee.minus(estimatedSellFee).abs()
+  const estimatedOasisFee = feeDiff.gt(new BigNumber(3))
```
piekczyk, line 324: So we won't show it as separate items in changes information list?

features/automation/optimization/controls/AutoBuyFormControl.tsx
```diff
@@ -212,9 +212,11 @@ export function AutoBuyFormControl({
   })
 
   const { debtDelta, collateralDelta } = getBasicBSVaultChange({
-    basicBSState: basicBuyState,
-    vault,
+    targetCollRatio: basicBuyState.targetCollRatio,
```
adamskrodzki, line 215: not sure why you split basicBuyState into needed properties, but maybe then to be consistent also split vault ?

features/automation/optimization/sidebars/SidebarSetupConstantMultiple.tsx
```diff
@@ -304,30 +312,44 @@ function ConstantMultipleInfoSectionControl({
...
-  // TODO: PK get those values from constantMultipleState when there is actual data
-  const targetColRatio = new BigNumber(200)
-  const multiplier = 2
+  // TODO: PK where do I get slippage?
   const slippage = new BigNumber(0.5)
-  const triggerColRatioToBuy = new BigNumber(200)
-  const triggerColRatioToSell = new BigNumber(300)
-  const maxPriceToBuy = new BigNumber(1600)
-  const minPriceToSell = new BigNumber(1200)
+  const feeDiff = estimatedBuyFee.minus(estimatedSellFee).abs()
```
adamskrodzki, line 323: what is this magic here ?

features/automation/basicBuySell/InfoSections/ConstantMultipleInfoSection.tsx
```diff
@@ -53,13 +59,35 @@ export function ConstantMultipleInfoSection({
         },
         {
           label: t('constant-multiple.vault-changes.cost-per-adjustment'),
-          // TODO: PK calculate this value
-          value: '$0',
+          value:
+            estimatedGasPriceOnTrigger &&
+            estimatedOasisFee
+              .map((feeItem) => {
+                return `$${formatAmount(feeItem.plus(estimatedGasPriceOnTrigger), 'USD')}`
```
piekczyk, line 66: change `estimatedGasPriceOnTrigger` to `estimatedGasConstOnTrigger`

piekczyk (changes requested):

features/automation/optimization/sidebars/SidebarSetupConstantMultiple.tsx
piotrkonowrocki: This line is checking diff between auto buy and auto sell fee for current settings. Depending on that difference later I'm deciding if I can display aggregated estimated cost (if diff is small) or separated estimated fee as range (if diff is big).

features/automation/optimization/sidebars/SidebarSetupConstantMultiple.tsx
piotrkonowrocki: Yes, if diff between fees is greater than $3, we are showing range, if lesser, just the bigger value.

features/automation/optimization/sidebars/SidebarSetupConstantMultiple.tsx
adamskrodzki: This is agreed in designs i assume? Update magic  new BigNumber(3) into constant in some common config (probably same place where let say OASIS_FEE is stored)

features/automation/optimization/controls/AutoBuyFormControl.tsx
piotrkonowrocki: I split `basicBuyState` into dedicated properties because I needed to use this method also for constant mutliply, but constant multiply state is different, so I couldn't just pass it as whole. I changed this method so it's requiring only those props that are actually used inside of it.

But good point on that `vault`, this method doesn't need the whole state, I split it too.

features/automation/optimization/sidebars/SidebarSetupConstantMultiple.tsx
piotrkonowrocki: Agreed on sync call.

Took out magic number to helpers as const.

features/automation/optimization/sidebars/SidebarSetupConstantMultiple.tsx
piotrkonowrocki: Fixed.

piekczyk (approved):

adamskrodzki (approved):

PR #1345, opened 2022-08-09
Get CM trigger data

piotrkonowrocki:
# Get CM trigger data

Also covers parts from https://app.shortcut.com/oazo-apps/story/5220/constant-multiple-information-section.
  
## Changes 👷‍♀️

- Updated contract address and abi according to newest version,
- loaded CM trigger into code and displayed it into details section,
- extended `extractBasicBSData` method with `isInGroup` parameter and changed its props to object with interface for better readability.
  
## How to test 🧪

- May not be possible to test right now, because cache bot address is not updated and doesn't return trigger after it's created.


blockchain/abi/dummy-automation-bot-aggregator.json
```diff
@@ -1,390 +1,91 @@
 [
```
piekczyk, line 1: not sure if it still dummy

components/vault/OptimizationControl.tsx
```diff
@@ -110,7 +110,10 @@ export function OptimizationControl({
...
 
   const basicBuyData = automationTriggersData
-    ? extractBasicBSData(automationTriggersData, TriggerType.BasicBuy)
+    ? extractBasicBSData({
+        triggersData: automationTriggersData,
+        triggerType: TriggerType.BasicBuy,
+      })
     : undefined
 
   const vaultHasActiveTrigger = basicBuyData?.isTriggerEnabled
```
piekczyk, line 119: maybe not for now but it should be extended with check whether constant multiple is enabled

features/automation/common/basicBSTriggerData.ts
```diff
@@ -100,14 +106,25 @@ export function extractGroupTriggersData(
   }
 
   return {
-    [TriggerType.CMBasicBuy]: extractBasicBSData(groupData, TriggerType.BasicBuy),
-    [TriggerType.CMBasicSell]: extractBasicBSData(groupData, TriggerType.BasicSell),
+    [TriggerType.CMBasicBuy]: extractBasicBSData({
```
piekczyk, line 109: we no longer have CMBasicBuy or CmBasicSell trigger type, we should operate on BasicBuy / BasicSell (automation package to be updated I guess)

features/automation/common/basicBSTriggerData.ts
```diff
@@ -100,14 +106,25 @@ export function extractGroupTriggersData(
...
+      triggersData: groupData,
+      triggerType: TriggerType.BasicBuy,
```
piekczyk, line 111: shouldn't we pass here and below `isInGroup` as `true`?

features/automation/common/basicBSTriggerData.ts
```diff
@@ -100,14 +106,25 @@ export function extractGroupTriggersData(
...
+    }),
+    [TriggerType.CMBasicSell]: extractBasicBSData({
+      triggersData: groupData,
+      triggerType: TriggerType.BasicSell,
+    }),
   }
 }
```
piekczyk, line 118: ok, I guess this whole method is not used everywhere. If its true lets remove it.

piekczyk (changes requested):

components/vault/OptimizationControl.tsx
piotrkonowrocki: I plan to check all CM logic for things like that while working on default state, this PR is just about getting data so I can move on with next thing.

features/automation/common/basicBSTriggerData.ts
piotrkonowrocki: This method is used in `useConstantMultipleStateInitialization` which I intend to refactor and remove some old stuff with default state.

features/automation/common/basicBSTriggerData.ts
piotrkonowrocki: Same thing, this method is probably going to be removed, the only reason I updated is that it was required to work in current state and parameters format changed.

features/automation/common/basicBSTriggerData.ts
piotrkonowrocki: Currently is used in `useConstantMultipleStateInitialization`, will be removed later.

blockchain/abi/dummy-automation-bot-aggregator.json
piotrkonowrocki: Renamed, removed `dummy`.

piekczyk (approved):

PR #1364, opened 2022-08-16
Prevent uBlock from blocking mixpanel analytics

halaprix:
# [Prevent uBlock from blocking mixpanel analytics](https://app.shortcut.com/oazo-apps/story/5645/prevent-ublock-from-blocking-mixpanel-analytics)

  
## Changes 👷‍♀️
- remove double tracking (directly from FE + proxied) for FE events
- proxy all tracked events through mixpanel endpoint (`/api/t`)
- added `$current_url` , `$initial_referrer` and `$initial_referring_domain` as per `https://help.mixpanel.com/hc/en-us/articles/115004613766-Default-Properties-Collected-by-Mixpanel` to supplement the data forwarded by proxy
- moved ` mixpanelInit()` in `_app.tsx` to the `useEffect` so `window` object  is visible ( we can get current URL)
-  removed `be-` prefix to keep the current data format
- if the user opted out (cookie settings) track only `AccountChange` address and `ConfirmTransaction`
  
## How to test 🧪

- change `MIXPANEL_ENV` to `production`
- remove `if (process.env.NODE_ENV !== 'development') {` in `_app.tsx` to track route changes
- check if the calls to `/api/t` are successful
- check events mixpanel dashboard in `Testing Environments`
- check mixpanel events in terminal when `MIXPANEL_ENV=='developement'`

### DIFF
--- a/analytics/analytics.ts
+++ b/analytics/analytics.ts
@@ -6,6 +6,7 @@ import getConfig from 'next/config'
 export type MixpanelDevelopmentType = {
   track: (eventType: string, payload: any) => void
   get_distinct_id: () => string
+  has_opted_out_tracking: boolean
 }
 
 export function enableMixpanelDevelopmentMode<T>(mixpanel: T): T | MixpanelDevelopmentType {
@@ -17,6 +18,7 @@ export function enableMixpanelDevelopmentMode<T>(mixpanel: T): T | MixpanelDevel
         console.info('Mixpanel Event: ', eventType, payload)
       },
       get_distinct_id: () => 'test_id',
+      has_opted_out_tracking: false,
     }
   }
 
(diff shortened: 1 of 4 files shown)


analytics/analytics.ts
```diff
@@ -6,6 +6,8 @@ import getConfig from 'next/config'
 export type MixpanelDevelopmentType = {
   track: (eventType: string, payload: any) => void
   get_distinct_id: () => string
+  has_opted_out_tracking: () => boolean
+  get_property: (propertyName: string) => any
```
jakubswierczek, line 10: I prefer avoiding `any`. Also one case returns `null` so that also should be in typing.

analytics/analytics.ts
```diff
@@ -81,121 +127,150 @@ export const trackingEvents = {
       walletType,
     }
 
-    mixpanel.track(eventName, eventBody)
     mixpanelInternalAPI(eventName, eventBody)
   },
 
   searchToken: (
     page: Pages.LandingPage | Pages.OpenVaultOverview | Pages.VaultsOverview,
```
jakubswierczek, line 134: can we merge it in one union type?

analytics/analytics.ts
halaprix: ok! changed that. tbh previously I've just copied the typing directly from `mixpanel-browser` -> `export function get_property(property_name: string): any;` :)

analytics/analytics.ts
halaprix: I'd leave it as Rafal did it - with the direct use of enum - there are 38 event types, feels unnecessary to change that.

PR #1367, opened 2022-08-16
fixed reset button for constant multiple form

johnnieskywalker:
# [Range slider - reset button issue](https://app.shortcut.com/oazo-apps/story/5677/range-slider-reset-button-issue)

  
## Changes 👷‍♀️
  Added extra isResetAction to prevent recalculation slider values (dependent on multiple variables) when one hits reset button, to make it behave in more predictable way, reset the initial state
  
## How to test 🧪
- Have a vault with constant multiple trigger enabled
- view details of cm trigger, memorize initial state on slider values
- edit slider or multiplier
- hit reset button, it should bring slider to same state as when initially displayed

### DIFF
--- a/scripts/docker/launch.sh
+++ b/scripts/docker/launch.sh
@@ -12,4 +12,4 @@ docker-compose "$@" pull
 (sleep 10 && cd ../.. && DATABASE_URL="postgresql://user:pass@localhost:5432/db?schema=public" yarn migrate)&
 
 export DATABASE_URL="postgresql://user:pass@postgres-oasis-borrow:5432/db?schema=public"
-docker-compose -p oasis-borrow --env-file ../../.env "$@" up
+docker-compose -p oasis-borrow --env-file ../../.env --env-file ../../.env.local "$@" up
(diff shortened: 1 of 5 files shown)


scripts/docker/launch.sh
```diff
@@ -12,4 +12,4 @@ docker-compose "$@" pull
 (sleep 10 && cd ../.. && DATABASE_URL="postgresql://user:pass@localhost:5432/db?schema=public" yarn migrate)&
 
 export DATABASE_URL="postgresql://user:pass@postgres-oasis-borrow:5432/db?schema=public"
-docker-compose -p oasis-borrow --env-file ../../.env --env-file ../../.env.local "$@" up
+docker-compose -p oasis-borrow --env-file ../../.env "$@" up
```
piotrkonowrocki, line 15: I think that shouldn't be commited, right?

piotrkonowrocki (changes requested):
One more thing: new value `isResetAction` is not set in state on some default value in `useConstantMultipleStateInitialization.ts`, so it's `undefined` until user does some action.

scripts/docker/launch.sh
johnnieskywalker: I'm not sure it was preventing app to build locally, but may be just matter of my environment @jakubswierczek any objections to commit this change ?

jakubswierczek: no, you can commit ;)

piotrkonowrocki: Still, is is the best place to do that in a random commit completely unrelated to its name and purpose? I suggest opening separated PR, that would make it easier to read files and commits history in the future.

piotrkonowrocki (approved):

PR #1373, opened 2022-08-17
Constant Multiple on non-multiply vaults

piotrkonowrocki:
# Constant Multiple on non-multiply vaults

https://app.shortcut.com/oazo-apps/story/5704/cm-banner-on-borrow-vault
https://app.shortcut.com/oazo-apps/story/5675/bug-remove-constant-multiple-from-borrow-vaults
https://app.shortcut.com/oazo-apps/story/5726/update-copy-cm-banner
  
## Changes 👷‍♀️

Changed Constant Multiple banner on non-multiply vaults:
- different text,
- CTA doesn't open Constant Multiple form, instead it moves user to overview tab.

Constant Multiple enabled in multiply vault that was switched to borrow stays intact, but after user removes it he can't set it up anymore until they go back to mutliply interface.

image

Important note: in MVP we are only moving user to overview tab, later on we should update the button so it would also chose "Switch to multiply" option in dropdown.
  
## How to test 🧪

- Try opening Constant Multiple in both multiply and borrow vault.

### DIFF
--- a/features/automation/optimization/controls/ConstantMultipleDetailsLayout.tsx
+++ b/features/automation/optimization/controls/ConstantMultipleDetailsLayout.tsx
@@ -8,17 +8,21 @@ import { ContentCardTargetMultiple } from 'components/vault/detailsSection/Conte
 import { ContentCardTotalCostOfFeature } from 'components/vault/detailsSection/ContentCardTotalCostOfFeature'
 import { ContentCardTriggerColRatioToBuy } from 'components/vault/detailsSection/ContentCardTriggerColRatioToBuy'
 import { ContentCardTriggerColRatioToSell } from 'components/vault/detailsSection/ContentCardTriggerColRatioToSell'
+import { VaultViewMode } from 'components/vault/GeneralManageTabBar'
 import {
   AUTOMATION_CHANGE_FEATURE,
   AutomationChangeFeature,
 } from 'features/automation/protection/common/UITypes/AutomationFeatureChange'
+import { VaultType } from 'features/generalManageVault/vaultType'
 import { useUIChanges } from 'helpers/uiChangesHook'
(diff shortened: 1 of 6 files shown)


features/automation/optimization/controls/ConstantMultipleDetailsLayout.tsx
```diff
@@ -107,12 +122,20 @@ export function ConstantMultipleDetailsLayout({
...
-                  type: 'Optimization',
-                  currentOptimizationFeature: 'constantMultiple',
-                })
+                if (isMultiplyVault) {
+                  uiChanges.publish(AUTOMATION_CHANGE_FEATURE, {
+                    type: 'Optimization',
+                    currentOptimizationFeature: 'constantMultiple',
+                  })
+                } else {
+                  setHash(VaultViewMode.Overview)
```
piekczyk, line 131: maybe beside that we should also switch form state to 'Switch to multiply'?

piekczyk (changes requested):

johnnieskywalker (approved):
Removed triiger that was enabled on borrow vault
https://goerli.etherscan.io/tx/0xf04500293f743cfe5763d88f77f310863fc99912e9d9b11df72532beeef999a8

after doing so expected banner appeared 
image

redirects to overview 
image

features/automation/optimization/controls/ConstantMultipleDetailsLayout.tsx
piotrkonowrocki: Let me quote myself from the PR description 😉 : "Important note: in MVP we are only moving user to overview tab, later on we should update the button so it would also chose "Switch to multiply" option in dropdown."

piekczyk (approved):

PR #1397, opened 2022-08-24
Opening position stETH

jakubswierczek:


features/earn/aave/common/state/types.ts
```diff
@@ -0,0 +1,67 @@
...
+      readonly type: 'PROXY_ADDRESS_RECEIVED'
+      readonly proxyAddress: string | undefined
+    }
+  | {
+      readonly type: 'CREATE_PROXY'
+    }
+  | {
+      readonly type: 'PROXY_CREATED'
+      readonly proxyAddress: string
+    }
```
paszkowskiDamian, line 16: Shouldn't this type be somewhere where proxy machine exists? It think it shouldn't be under aave scope.

jakubswierczek: You are right but these are events in aave machines. Proxy machine has its own

paszkowskiDamian (approved):

PR #1398, opened 2022-08-24
Adjusted auto sell immediately validation

piekczyk:
# [Adjusted auto sell immediately validation](https://shortcutLink)

## Changes 👷‍♀️

- immediately validation in Auto Sell and Constant Multiply form now takes into account current vault state
- added validation to Constant Multiply regarding dust limit
  
## How to test 🧪

- try to setup (both in Auto Sell form and in CM form) trigger that would hit dust limit at current vault state or in future (error message should be visible)
- warning regarding immediately execution in both form should be displayed only if trigger is eligible based on current vault state

### DIFF
--- a/features/automation/optimization/validators.ts
+++ b/features/automation/optimization/validators.ts
@@ -95,16 +95,20 @@ export function warningsConstantMultipleValidation({
   gasEstimationUsd,
   ethBalance,
   ethPrice,
+  debtFloor,
   sliderMin,
   isStopLossEnabled,
   isAutoBuyEnabled,
   isAutoSellEnabled,
   constantMultipleState,
+  debtDeltaWhenSellAtCurrentCollRatio,
 }: {
   vault: Vault
   ethBalance: BigNumber
+  debtFloor: BigNumber
   ethPrice: BigNumber
   sliderMin: BigNumber
+  debtDeltaWhenSellAtCurrentCollRatio: BigNumber
   gasEstimationUsd?: BigNumber
   isStopLossEnabled: boolean
   isAutoBuyEnabled: boolean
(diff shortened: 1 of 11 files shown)


features/automation/optimization/validators.ts
```diff
@@ -130,9 +134,9 @@ export function warningsConstantMultipleValidation({
 
   const addingConstantMultipleWhenAutoSellOrBuyEnabled = isAutoBuyEnabled || isAutoSellEnabled
 
-  const constantMultipleAutoSellTriggeredImmediately = sellExecutionCollRatio
-    .div(100)
-    .gte(vault.collateralizationRatioAtNextPrice)
+  const constantMultipleAutoSellTriggeredImmediately =
+    sellExecutionCollRatio.div(100).gte(vault.collateralizationRatioAtNextPrice) &&
+    !debtFloor.gt(vault.debt.plus(debtDeltaWhenSellAtCurrentCollRatio))
```
adamskrodzki, line 139: this is added so you do not display error and warning at the same time?

features/automation/protection/controls/AutoSellFormControl.tsx
```diff
@@ -200,6 +200,13 @@ export function AutoSellFormControl({
     collateral: vault.lockedCollateral,
     vaultDebt: vault.debt,
   })
+
+  const executionPriceAtCurrentCollRatio = collateralPriceAtRatio({
```
adamskrodzki, line 204: either strange naming or error 

` collateralPriceAtRatio`

sounds like you try to compute price that would fullfill conditions passed as parameters,

but then `colRatio: vault.collateralizationRatio,` looks like passing current coll ratio, so what you should get back is current oracle price

why then just not take it directly. Also suspicious since BB/BS/CM  operates on next price

piekczyk: yep, I could take it directly without computing. I will update it. 

I used here current `collaterizationRatio` because if added trigger will be eligible for execution immediately we need debt change from its current state (you can compare it to manual decrease of multiply be the user) not from state from next price update.

features/automation/optimization/validators.ts
piekczyk: We are showing errors and warnings independently. I extended this condition to make sure that this warning will be visible only if trigger execution will be eligible and immediate

piotrkonowrocki (approved):

PR #1403, opened 2022-08-25
new default states for ab and as, sl

johnnieskywalker:
# [Prepare default states for SL and AS/AB](https://app.shortcut.com/oazo-apps/story/5781/prepare-default-states-for-sl-and-as-ab)

## How to test 🧪
  
- Auto-Buy, Auto-Sell, Stop-Loss should have new default states if trigger doesn't exist
- if trigger exist than default state should be loaded from trigger (no regression)
- reset buttons for vaults with triggers resets data to the same as on existing trigger
- reset button for vaults without trigger, brings back form to default state for vaults without trigger as defined in task
Test for edge cases like vaults with low amount of collateral etc. to figure values for such edge cases


piekczyk:
I found regression following regression:

1) if trigger exists and you visit protection tab, after pills are visible immediately (isEditing is true)
image

2) if you try to remove existing trigger stop loss coll ratio in summary is broken and additionally if you click on `Return to adjusting Stop-Loss` app crash
image

after clicking on `Return to adjusting Stop-Loss`:
image

piekczyk:
When clicking on `Reset` dots switches their places. It may be potentially fixed if you merge dev.
image

piekczyk:
When clicking on reset after trigger removal nothing happens

image

piekczyk:
when stop loss is set up close to maxium a default state after refresh push it out of range (after pills visible without editing)
image

piekczyk:
**EDGE CASE** when stop loss is set up close to max value and you try to enable auto buy, default value is set up out of auto buy range (range 352 - 500, default set at 350)
image

features/automation/protection/controls/sidebar/SidebarAdjustStopLossEditingStage.tsx
```diff
@@ -101,7 +117,11 @@ export function SidebarAdjustStopLossEditingStage({
                 })
                 uiChanges.publish(ADD_FORM_CHANGE, {
                   type: 'stop-loss',
-                  stopLoss: stopLossLevel.times(100),
+                  stopLoss: stopLossLevel.isZero()
```
johnnieskywalker, line 120: Behaviour of reset after this commit 
image

features/automation/protection/common/stopLossTriggerData.ts
```diff
@@ -74,14 +74,19 @@ export function prepareStopLossTriggerData(
...
+
   return {
     cdpId: vaultData.id,
     triggerType,
     proxyAddress: vaultData.owner,
     triggerData: encodeTriggerDataByType(CommandContractType.CloseCommand, [
       vaultData.id.toString(),
       triggerType.toString(),
-      stopLossLevel.toString(),
+      stopLossLevel.gte(100) ? stopLossLevel.toString() : stopLossLevel.times(100).toString(),
```
johnnieskywalker, line 89: probably not the best way to fix. Fixes when returning to adjust, but still value passed to UI remains wrong

features/automation/protection/useStopLossStateInitializator.ts
```diff
@@ -34,21 +34,17 @@ export function useStopLossStateInitializator(
...
     initialStopLossSelected: selectedStopLossCollRatioIfTriggerDoesntExist,
-  })
+  }).multipliedBy(100)
 
   useEffect(() => {
     uiChanges.publish(ADD_FORM_CHANGE, {
       type: 'close-type',
       toCollateral: isToCollateral,
     })
-    uiChanges.publish(ADD_FORM_CHANGE, {
```
johnnieskywalker, line 42: that was causing err when returning to adjust

johnnieskywalker:
> When clicking on `Reset` dots switches their places. It may be potentially fixed if you merge dev. image

it fixed the issue 
image

components/vault/VaultChangesInformation.tsx
```diff
@@ -153,7 +153,7 @@ export function getEstimatedGasFeeText(gasEstimation?: GasEstimationContext, wit
     case undefined:
       return <EstimationError withBrackets={withBrackets} />
     case GasEstimationStatus.calculated:
-      const textGas = `$${formatAmount(gasEstimation.usdValue, 'USD')}`
+      const textGas = `$${formatAmount(gasEstimation?.usdValue, 'USD')}`
```
johnnieskywalker, line 156: Fix works, sometimes usdValue was undefined, which was not checked and it caused an error
image

johnnieskywalker:
> **EDGE CASE:** when stop loss is set up close to maxium a default state after refresh push it out of range (after pills visible without editing) image

issue seems to be gone but plz test

johnnieskywalker:
> **EDGE CASE** when stop loss is set up close to max value and you try to enable auto buy, default value is set up out of auto buy range (range 352 - 500, default set at 350). The same applies to auto sell form. image
> 
> image

Probably in such case none of values fo ab/as make sense
image

This is handled in CM and will create another story to handle it in ab/as

image

halaprix (changes requested):
image
stop loss button disabled by default and after resetting

halaprix (approved):
The button works as intended. It goes back to the default state - which is disabled.

PR #1422, opened 2022-09-06
handled too high c-ratio for ab

johnnieskywalker:
# [When SL is close to max and/or c-reatio above 500 AB/AS show up with illegal values by default](https://app.shortcut.com/oazo-apps/story/5935/when-sl-is-close-to-max-ab-as-show-up-with-illegal-values-by-default)

  
## Changes 👷‍♀️
- item 1
  
## How to test 🧪
- Case coll ratio too high in AB:

image

click redirects, if c-ratio is lower than slider max user is able to set trigger
image

AB - check for sl too high
image
redirects to protection 
image

disappears after adjusting 
image

SL too high for AS example
image

coll ratio too high example
image

### DIFF
--- a/features/automation/optimization/sidebars/SidebarAutoBuyEditingStage.tsx
+++ b/features/automation/optimization/sidebars/SidebarAutoBuyEditingStage.tsx
@@ -91,32 +91,10 @@ export function SidebarAutoBuyEditingStage({
     .times(100)
     .gt(sliderMax)
 
-  if (isCurrentCollRatioHigherThanSliderMax) {
-    return (
-      <Trans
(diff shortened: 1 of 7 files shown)


features/automation/optimization/sidebars/SidebarAutoBuyEditingStage.tsx
```diff
@@ -222,6 +283,21 @@ export function SidebarAutoBuyEditingStage({
       )}
     </>
   )
+
+  function adjustDefaultValuesIfOutsideSlider() {
```
johnnieskywalker, line 287: it happened when for ex sl was set to 440 on vault 369
image

piekczyk (approved):

PR #1430, opened 2022-09-07
Automation context provider

piekczyk:
# [Automation context provider](https://app.shortcut.com/oazo-apps/story/5775/reduce-amount-of-calls-to-same-functions)

## Changes 👷‍♀️

- created automation context provider which is used on the top level of mange vault view. 
- automation context provider contains decoded triggers data per each automation feature so its calculated only once not many times in different places as before
  
## How to test 🧪

- do round of adding / updating / removing auto triggers, everything should work as intended as this PR is purely about refactor

### DIFF
--- a/components/vault/OptimizationControl.tsx
+++ b/components/vault/OptimizationControl.tsx
@@ -1,11 +1,9 @@
 import { Icon } from '@makerdao/dai-ui-icons'
-import { TriggerType } from '@oasisdex/automation'
 import { IlkData } from 'blockchain/ilks'
 import { Vault } from 'blockchain/vaults'
 import { useAppContext } from 'components/AppContextProvider'
+import { useAutomationContext } from 'components/AutomationContextProvider'
 import { AppLink } from 'components/Links'
-import { extractBasicBSData } from 'features/automation/common/basicBSTriggerData'
-import { extractConstantMultipleData } from 'features/automation/optimization/common/constantMultipleTriggerData'
 import { OptimizationDetailsControl } from 'features/automation/optimization/controls/OptimizationDetailsControl'
 import { OptimizationFormControl } from 'features/automation/optimization/controls/OptimizationFormControl'
 import { VaultType } from 'features/generalManageVault/vaultType'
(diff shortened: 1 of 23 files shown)


components/vault/OptimizationControl.tsx
```diff
@@ -108,28 +106,17 @@ export function OptimizationControl({
...
-        triggersData: automationTriggersData,
-        triggerType: TriggerType.BasicBuy,
-      })
-    : undefined
-  const constantMultipleTriggerData = automationTriggersData
-    ? extractConstantMultipleData(automationTriggersData)
-    : undefined
-
-  const vaultHasActiveAutoBuyTrigger = basicBuyData?.isTriggerEnabled
-  const vaultHasActiveConstantMultipleTrigger = constantMultipleTriggerData?.isTriggerEnabled
```
adamskrodzki, line 132: actually why `basicBuyData`/`constantMultipleTriggerData` etc even can be null? under what circumstances it is not null and yet `isTriggerEnabled` is false ?

components/vault/ProtectionControl.tsx
```diff
@@ -104,66 +102,39 @@ function getZeroDebtProtectionBannerProps({
...
-    : undefined
-  const basicSellData = automationTriggersData
-    ? extractBasicBSData({
-        triggersData: automationTriggersData,
-        triggerType: TriggerType.BasicSell,
-      })
-    : undefined
-  const vaultHasActiveTrigger = stopLossData?.isStopLossEnabled || basicSellData?.isTriggerEnabled
+  const vaultHasActiveTrigger =
+    stopLossTriggerData.isStopLossEnabled || autoSellTriggerData.isTriggerEnabled
```
adamskrodzki, line 118: looks inconsistent with OptimisationControl, why here you do not check for nulls ? comment from ProtectionControl still holds.

features/automation/protection/triggers/AutomationTriggersData.ts
```diff
@@ -71,25 +70,21 @@ export function createAutomationTriggersChange$(
...
-          }),
-          basicBuyData: extractBasicBSData({
-            triggersData: triggers,
-            triggerType: TriggerType.BasicBuy,
-          }),
-          constantMultipleData: extractConstantMultipleData(triggers),
-        })),
-      )
-    : []
+  return automationTriggersData$(id).pipe(
```
adamskrodzki, line 73: we ignore feature toggle now ?

PR #1444, opened 2022-09-09
Dedicated method for automation sidebar copies

piekczyk:
# [Dedicated method for automation sidebar copies](https://app.shortcut.com/oazo-apps/story/5774/dedicated-methods-for-handle-automation-sidebar-copies)

## Changes 👷‍♀️

- prepared dedicated method for handling copies within automation features sidebars
- adjusted old methods which will stay for other features (removed automation flows handling from them)
  
## How to test 🧪

- test all flows within automation scope and pay attention to copies inside sidebar
- test stop loss setup during opening flow
- few test with opening / managing borrow and multiply vault to make sure that other copies are not broken


features/automation/common/AddAndRemoveTriggerControl.tsx
```diff
@@ -65,7 +72,7 @@ export function AddAndRemoveTriggerControl({
     [removeAllowance, proxyAddress, triggersId],
   )
 
-  function txHandler() {
+  function txHandler(options?: AddAndRemoveTxHandler) {
```
piotrkonowrocki, line 75: Is there a reason to use some `options` var as proxy holding whole interface? Wouldn't be it easier just to use spread?
```
function txHandler({callOnSuccess}: AddAndRemoveTxHandler) {
```

features/automation/common/getAutomationPrimaryButtonLabel.ts
```diff
@@ -0,0 +1,107 @@
...
+    case 'editBasicBuy':
+    case 'editConstantMultiple':
+      return 'automation.updating'
+    case 'cancelBasicSell':
+    case 'cancelBasicBuy':
+    case 'cancelConstantMultiple':
+    case 'cancelSl':
+      return 'automation.cancelling'
+    default:
+      return UNREACHABLE_CASE_MESSAGE
```
piotrkonowrocki, line 54: Why does other methods are throwing exception, but here (and in method below) there is a empty string as default return?

I don't remember if it's somehow justified and it was like that already in older methods? If so, I suggest moving `UNREACHABLE_CASE_MESSAGE` to 'helpers/UnreachableCaseError' and importing it in similar way like `UnreachableCaseError`.

piotrkonowrocki, line 54: Why does other methods are throwing exception, but here there is a empty string as default return?

I don't remember if it's somehow justified and it was like that already in older methods? If so, I suggest moving `UNREACHABLE_CASE_MESSAGE` to 'helpers/UnreachableCaseError' and importing it in similar way like `UnreachableCaseError`.

Same goes for line `72` below.

features/sidebar/getSidebarStatus.ts
```diff
@@ -18,25 +18,25 @@ function getSidebarProgressTxInProgressKey({
...
-      return 'auto-buy.cancelling-auto-buy'
-    case 'addBasicBuy':
-    case 'editBasicBuy':
-      return 'auto-buy.setting-auto-buy'
-    case 'cancelConstantMultiple':
-      return 'constant-multiple.cancelling-constant-multiple'
-    case 'addConstantMultiple':
-    case 'editConstantMultiple':
-      return 'constant-multiple.setting-constant-multiple'
+    // case 'adjustSl':
```
piotrkonowrocki, line 22: Why commented out instead of removing? Same for `52` line in the same file.

piotrkonowrocki (changes requested):

features/automation/common/AddAndRemoveTriggerControl.tsx
piekczyk: I tried to do it this way but typescript does not allow for object destructuring when parameter is optional

piotrkonowrocki (approved):

PR #1464, opened 2022-09-13
Remove Constant Multiple banner from non-multiply vaults

piotrkonowrocki:
# Remove Constant Multiple banner from non-multiply vaults
  
## Changes 👷‍♀️

As in title.
  
## How to test 🧪

See a protection tab in both borrow and multiply vaults and confirm if banner is visible. If Constant Multiple was enabled on multiply vault and then vault was transformed into borrow, Constant Multiple UI should still be available.


George-field (approved):
Nice

PR #1465, opened 2022-09-13
Lw/5781 default state sl open flow

johnnieskywalker:
# [Added initial state of sl when opening vault(s)](https://app.shortcut.com/oazo-apps/story/6019/add-default-state-to-stop-loss-during-vault-creation-flow)

  
## Changes 👷‍♀️
-added default sl selected for opening borrow and multiply vault
  
## How to test 🧪
Try to open vault, generate DAI and see if initial values of sl selected will make sense
- Test for multiply vault 
image

- Test for borrow vault
image


George-field (approved):
Looks good to me!

PR #1470, opened 2022-09-14
Disable access to Auto-sell and Auto-buy UI when Constant Multiple is enabled

piotrkonowrocki:
# [Disable access to Auto-sell and Auto-buy UI when Constant Multiple is enabled](https://app.shortcut.com/oazo-apps/story/6042/disable-auto-buy-and-auto-sell-banners-when-user-has-active-constant-multiple)

## Changes 👷‍♀️

- Updated `Banner` component so it accepts more than one paragraph of text with nice spacing between then,
- updated `getAutoFeaturesSidebarDropdown` method to not display Auto-buy and Auto-sell options if Constant Multiple is enabled, also do not display any dropdown if it would have only one item,
- disabled CTA buttons in Auto-buy and Auto-sell banners and added explanation messages to them why they are disabled.

image
  
## How to test 🧪

Test on a vault with and without Constant Multiple enabled if you can get access to Auto-sell and Auto-buy UI.

### DIFF
--- a/components/Banner.tsx
+++ b/components/Banner.tsx
@@ -29,8 +29,6 @@ type BannerProps = {
 
 export function Banner({ title, description, button, image, sx }: BannerProps) {
   const descriptionsArray = Array.isArray(description) ? description : [description]
-  console.log(description)
-  console.log(descriptionsArray)
 
   return (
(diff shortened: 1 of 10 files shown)


johnnieskywalker (approved):
image

image

PR #1489, opened 2022-09-26
remove community manager role

FBrinkkemper:
removed community manager role from carreers.


marcinciarka (approved):
Checked, community manager role is no more.
image

PR #1492, opened 2022-09-27
Gf/pnl card update

George-field:
# [Update PnL modal](https://app.shortcut.com/oazo-apps/story/5357/update-pnl-popup-card)

## Changes 👷‍♀️
  - Updated the PnL modal as per designs
  - Mobile view
  - added new values to PnL
  
## How to test 🧪
- head to a multiply vault that has transactions
- go to the overview page and click the Net Value section
- The modal will appear, check that the values are correct

Vault 392 on staging has a vault with history.

Please can we double check my calculations as well to make sure they are as expected.


.env
```diff
@@ -25,7 +25,7 @@ OPERATION_EXECUTOR_TEMP=0x3C1Cb427D20F15563aDa8C249E71db76d7183B6c
 EXCHANGE=0xb5eB8cB6cED6b6f8E13bcD502fb489Db4a726C7B
 DUMMY_EXCHANGE=0x7bc06c482DEAd17c0e297aFbC32f6e63d3846650
 USE_DUMMY=0
-BLOCK_NUMBER=14709959
+BLOCK_NUMBER=15616946
```
piotrkonowrocki, line 28: Not sure if we need to commit things like that? I assume we don't.

components/vault/detailsSection/ContentCardNetValue.tsx
```diff
@@ -64,49 +78,51 @@ function ContentCardNetValueModal({
...
-        ) : (
-          <Box />
-        )}
-        <Box>{debtUSD}</Box>
+        <Box>
+          <Text
+            sx={{ fontWeight: 'semiBold' }}
+          >{`${netValueUndercollateralizedToken} ${token}`}</Text>
+          <Text sx={{ fontSize: 1, color: 'neutral80', fontWeight: 'semiBold' }}>
+            Total Col: {`${lockedCollateral} ${token}`}
```
piotrkonowrocki, line 116: Missing translation.

components/vault/detailsSection/ContentCardNetValue.tsx
```diff
@@ -64,49 +78,51 @@ function ContentCardNetValueModal({
...
-        {renderCollateralValue ? (
-          <Box>{`${netValueUndercollateralizedToken} ${token}`}</Box>
-        ) : (
-          <Box />
-        )}
-        <Box>{netValueUSD}</Box>
+        <Box>
+          <Text sx={{ fontWeight: 'semiBold' }}>{netValueUSD}</Text>
+          <Text sx={{ fontSize: 1, color: 'neutral80', fontWeight: 'semiBold' }}>
+            Debt: {lockedCollateralUSD}
```
piotrkonowrocki, line 123: Missing translation.

components/vault/detailsSection/ContentCardNetValue.tsx
```diff
@@ -219,6 +273,7 @@ export function ContentCardNetValue({
       precision: 2,
       roundMode: BigNumber.ROUND_DOWN,
     })}`,
+    currentPnLValue: ``,
```
piotrkonowrocki, line 276: Why empty template literal instead of just empty string?

components/vault/detailsSection/ContentCardNetValue.tsx
```diff
@@ -255,6 +320,14 @@ export function ContentCardNetValue({
     debtTokenMarketPrice: formatted.debtTokenMarketPrice,
     netValueOraclePrice: formatted.netValueOraclePrice,
     netValueMarketPrice: formatted.netValueMarketPrice,
+    depositTotalAmounts: {
```
piotrkonowrocki, line 323: Wouldn't
```
depositTotalAmounts = formatted.depositTotalAmounts
```
be shorter? Or am I missing something?

features/multiply/manage/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+  const validTypes = type === 'DEPOSIT' ? ['DEPOSIT', 'OPEN'] : ['WITHDRAW']
+  const events = historyEvents.filter((event) => validTypes.includes(event.kind))
+
+  let totalDolarAmount = new BigNumber(0)
+  let totalEthAmount = new BigNumber(0)
+
+  for (const event of events) {
+    if (type === 'WITHDRAW') {
+      totalDolarAmount = totalDolarAmount.plus(event.daiAmount || 0)
+      console.log(event.collateralAmount?.toString())
```
piotrkonowrocki, line 24: `console` left.

features/multiply/manage/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Function to calculate the sum of gas fees for vault history events.
+ * @param historyEvents
+ * @returns BigNumber
+ */
+export function calculateTotalGasFeeInEth(historyEvents: VaultHistoryEvent[]) {
+  let totalGasFees = new BigNumber(0)
+
+  if (historyEvents) {
+    for (const event of historyEvents) {
+      console.log(event && event.gasFee?.toString())
```
piotrkonowrocki, line 49: `console`

components/vault/detailsSection/ContentCardNetValue.tsx
```diff
@@ -160,25 +170,55 @@ function ContentCardNetValueModal({
...
+      </Box>
+      <Grid
+        sx={{ textAlign: 'right', gridTemplateColumns: 'repeat(3,1fr)' }}
+        gap={2}
+        columns={[1, 2, 3]}
+      >
+        <Box sx={{ fontSize: 1, fontWeight: 'medium', color: 'neutral80' }}>
+          {t('manage-multiply-vault.card.deposits')}
+        </Box>
+        <Text
```
piotrkonowrocki, line 184: General question: `depositTotalAmounts` is optional, is there a scenario where it is not provided to the component? Maybe it's worth to do couple ifs to skip entire `HTML` block if values are unavailable instead of displaying empty rows?

piotrkonowrocki (changes requested):
Additional note: I know it's not in the scope of that task, but maybe that's good moment to sort a little bit `HTML` semantics mess that we have in some places? I mean to add `as` prop to all `Box` and `Text` components that are missing it, so we can skip a `div` that should be `p` or `span`?

piotrkonowrocki:
I have a vault `210` on Goerli which I used for testing. That vault has long history of doing different things on it, so I was trying to check calculations:
image
Why does Deposits and Withdraws are worth $0 in dollars? And also we are displaying total gas fees in ETH?

George-field:
> I have a vault `210` on Goerli which I used for testing. That vault has long history of doing different things on it, so I was trying to check calculations: image Why does Deposits and Withdraws are worth $0 in dollars? And also we are displaying total gas fees in ETH?

In reference to your point about the gas fee in eth, I mentioned it to sam and he asked that it be kept. And I will check this vault to debug

components/vault/detailsSection/ContentCardNetValue.tsx
George-field: This is a mistake, shouldn't have been kept in.

components/vault/detailsSection/ContentCardNetValue.tsx
George-field: Yeah, Think I have used the spread operator too much in my life 🤦

components/vault/detailsSection/ContentCardNetValue.tsx
piotrkonowrocki: Okay, understood. If you find a way to solve that, that's great, if not and empty rows needs to be there so it won't break down - feel free to resolve this comment.

piotrkonowrocki:
It looks like issue with zeros were resolved:
image
Still not sure about rest of the values, can you confirm that this is correct? Deposit 4ETH and withdraw -4ETH? I'm not sure if that is right value in the context of this vault, even at this moment total collateral is 16+ETH. Or maybe I just don't get how is that suppose to work, if so, please a word of explanation 😄
image


Also, another tiny UI issue, withdraws are formatted differently than the rest, no zeros after comma.
image

features/multiply/manage/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+    for (const event of historyEvents) {
+      totalGasFees = totalGasFees.plus(event.gasFee || 0)
+    }
+  }
+
+  return amountFromWei(totalGasFees)
+}
+
+export function calculateCurrentPnLInUSD(pnl: BigNumber, netValueUSD: BigNumber) {
+  return pnl.times(netValueUSD).dividedBy(pnl.plus(1))
```
adamskrodzki, line 83: let say pnl = 0.5 (user made +50% of initial investment) and netValueUSD aquals 1500 expected value is 500, right ?

0.5* 1500 / (0.5+1) == 500.

Seems ok

adamskrodzki (approved):

piotrkonowrocki (dismissed):

PR #1503, opened 2022-09-30
Gnosis Safe Dirty Fix

BackToTheCode:
# [Ticket in Shortcut](https://app.shortcut.com/oazo-apps/story/6134/multiple-vault-gas-estimates-failing-w-gnosis-safe-and-not-sending-t-x)
  
## Changes 👷‍♀️
- Skip gas estimation for specific multiply vault actions with gnosis safe
  
## How to test 🧪
- Create Gnosis Safe
- Go to gnosis safe on mainnet or goerli
- Open an ETH-A multiply vault or manage existing
- Withdraw some ETH 
- Also try and close vault

Expect gas estimation to fail (n/a) but for transaction to be sent correctly by gnosis provider. Should see popup modal and correct gas estimation in there.


paszkowskiDamian (approved):

features/multiply/manage/pipes/manageMultiplyVaultTransactions.ts
```diff
@@ -647,6 +659,14 @@ export function applyEstimateGas(
...
     } = state
 
+    /*
+    For Gnosis wallets we experienced gas estimation errors for certain actions.
+    This errors occurred within the Gnosis SDK are proved tricky to debug. Given the Gnosis t/x confirmation popup does correctly estimate gas
+    We've opted to simply disable gas estimation on our side for those wallets for select actions
+    */
+    const isGnosisSafeWallet = context.connectionKind === 'gnosisSafe'
+    const GNOSIS_GAS_ESTIMATE = undefined
+
```
paszkowskiDamian, line 669: One note, I guess you could short circuit it here and just return  `GNOSIS_GAS_ESTIMATE` whenever it's gnosisSafe connection then you wouldn't need to mess with the logic down there.

paszkowskiDamian (approved):

PR #1504, opened 2022-10-03
Auto-take Profit min/max on slider

piotrkonowrocki:
# Auto-take Profit min/max on slider

Min value on slider: `current collateral price * 1.05`,
Max value on slider: `all time high collateral price * 2`, or, if API is not responding, `current collateral price * 10`,
Default value on slider: price of collateral based on `current collateral ratio * 1.2`.

image
  
## Changes 👷‍♀️

- created API endpoint to request for all time high price from coingecko,
- added slider min and max values to `getAutoTakeProfitStatus` method,
- forced Auto-take Profit slider to be based on potential price, not collateral ratio,
- updated `AUTO_TAKE_PROFIT_FORM_CHANGE` reduced with method that allows to set up default values and also removed `execution-coll-ratio` event and replaced it with `execution-price` since contract requires price, not col. ratio.
  
## How to test 🧪

Check out if slider min, max and default values on Auto-take Profit are working in a way I described them above.

### DIFF
--- a/blockchain/tokensMetadata.ts
+++ b/blockchain/tokensMetadata.ts
@@ -19,6 +19,7 @@ export interface TokenConfig {
   token0?: string
   token1?: string
   coinbaseTicker?: string
+  coinGeckoId?: string
   background?: string
   digitsInstant?: number
 }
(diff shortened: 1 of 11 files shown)


johnnieskywalker (approved):
Everything as expected, no comments on code, nice way of handling aths
image

I know there's no ATH data for LP tokens. I'm bit not sure about BAT-A

PR #1507, opened 2022-10-04
Discover oasis page

piotrkonowrocki:
# Discover oasis page

I decided that instead of creating one `/discovery` page, from performance and SEO point of view, it would be better to create four different ones  for each type of table. For now I made only one page for "High risk positions", but main template already assumes that it is going to be used to create next ones.

If user goes directly to `/discovery`, he is going to be redirected to `/discovery/high-risk-positions` as it first options in navigation.

image
  
## Changes 👷‍♀️

- created new feature flag for Discovery: `DiscoverOasis`,
- updated main app navigation, tidied it a little bit, but without huge revolution since it looks like we might be working on bigger navigation redesign in near future (see: https://discord.com/channels/837076147694207067/839057646480785438/1026803561079787550),
- created empty page for Discovery pages,
  
## How to test 🧪

- Enable `DiscoverOasis` feature flag,
- compare if main navigation has "Discovery" only when feature flag is enabled,
- go to `/discovery` page to check if redirect works properly.

Discovery page itself for now is just an intro and lorem ipsum, so there is nothing to test inside of it.

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -442,14 +442,13 @@ const LINKS = {
   multiply: `/multiply`,
   borrow: `/borrow`,
   earn: '/earn',
-  discovery: '/discovery/high-risk-positions',
+  discovery: '/discovery',
 }
 
 function ConnectedHeader() {
   const { uiChanges } = useAppContext()
-  const { pathname } = useRouter()
-  const { t } = useTranslation()
   const onMobile = useOnMobile()
+
   const [widgetUiChanges] = useObservable(
     uiChanges.subscribe<SwapWidgetState>(SWAP_WIDGET_CHANGE_SUBJECT),
   )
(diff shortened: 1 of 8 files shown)


components/Header.tsx
```diff
@@ -442,21 +442,21 @@ const LINKS = {
   multiply: `/multiply`,
   borrow: `/borrow`,
   earn: '/earn',
+  discovery: '/discovery',
```
adamskrodzki, line 445: I know I'm picky, but all others are verbs/calls to action so I guess `/discover` not `/discovery`

johnnieskywalker (approved):
Approved as expected, just not sure if navbar text should be Discovery or Discover but this is just a detail
image

components/Header.tsx
```diff
@@ -555,7 +518,121 @@ function ConnectedHeader() {
...
+      as="li"
+      sx={{ position: 'relative' }}
+      onMouseEnter={() => {
+        setIsMouseOver(true)
+      }}
+      onMouseLeave={() => {
+        setIsMouseOver(false)
+      }}
+    >
+      {link ? (
```
adamskrodzki, line 551: So if(link) we ignore chldren ?

adamskrodzki: How about writing tests for this component :D ?

components/Header.tsx
```diff
@@ -868,6 +921,79 @@ function DisconnectedHeader() {
...
+  return (
+    <Flex
+      sx={{
+        alignItems: 'center',
+        justifyContent: ['space-between', 'flex-start'],
+        width: ['100%', 'auto'],
+      }}
+    >
+      <Logo />
+      {!discoverOasisEnabled ? (
```
adamskrodzki, line 939: is it everywhere like that with feature toggle ? 

`!toggle`<?(...):(....)

if no then easier to red is

  `toggle`?(...):(....)

so change order of controls

components/Header.tsx
piotrkonowrocki: I struggled with that too, but it is like that in the figma, so I assumed that name is intentional.

components/Header.tsx
piotrkonowrocki: For now on I assumed that item in navigation can either be a link or dropdown, but not both thing at once. You thing that assumption is incorrect? There is no case in the figma for nav item that is both of those things.

components/Header.tsx
piotrkonowrocki: I don't know if it is like that everywhere, but that is how it suppose to be and how I was implementing that. If it's condition with else as well. First there should be default state, which is a state when feature is disabled, then alternate version, so when feature is enabled. And because of the way our feature toggles are made, it has to be with negation in the beginning.

components/Header.tsx
johnnieskywalker: me too lol ;d

PR #1522, opened 2022-10-07
Generic dropdown component

piotrkonowrocki:
# Generic dropdown component

Partially covers https://app.shortcut.com/oazo-apps/story/6101/tbc-discover-oasis-filters-ui-component.

Based on [react-select](https://react-select.com), since it was already in codebase and I didn't wanted to duplicate similar solutions. Dropdown for Discovery Oasis, but with made in more generic way that will allow use it anywhere else where needed. By default it looks like one from Discovery, but all style options can be overwritten freely with `customStyles` prop and following instructions from documentation: https://react-select.com/styles.

image
  
## Changes 👷‍♀️

- Created `GenericSelect` component.
  
## How to test 🧪

This is only the component without any usage and examples. To test it use following snippets on empty page:

### Data sources
```
  const ranges = [
    { value: '>100k', label: 'Over $100' },
    { value: '75k-100k', label: '$75,000 - $100,000' },
    { value: '50k-75k', label: '$50,000 - $75,000' },
    { value: '<50k', label: 'Below $50,000' },
  ]
  const tokens = [
    { value: 'all', label: 'All Asset' },
    { value: 'eth', label: 'ETH', icon: getToken('ETH').iconCircle },
    { value: 'dai', label: 'DAI', icon: getToken('DAI').iconCircle },
    { value: 'wbtc', label: 'WBTC', icon: getToken('WBTC').iconCircle },
  ]
```

### Basic example
Just a dropdown.
```
<GenericSelect options={ranges} />
```

### More advanced example
Contains options with icons, example of custom styling and doesn't display selected position on the list.
```
<GenericSelect
  options={tokens}
  defaultValue={tokens[0]}
  customStyles={{
    option: ({ isSelected }) => ({
      ':hover': {
        backgroundColor: 'red',
      },
      ...(isSelected && { display: 'none' }),
    }),
  }}
/>
```


components/GenericSelect.tsx
```diff
@@ -0,0 +1,202 @@
...
+      {},
+    )
+
+  return (
+    <Box sx={{ position: 'relative' }}>
+      <ReactSelect
+        blurInputOnSelect={true}
+        isDisabled={isDisabled}
+        isSearchable={isSearchable}
+        menuIsOpen={true}
```
piekczyk, line 142: please add here short comment with explanation why it is hardcoded to `true` so we avoid situation that at some point one of our FE devs will 'fix' it without knowing the whole context

piekczyk (changes requested):

PR #1523, opened 2022-10-08
Discover Oasis Filters UI Component

piotrkonowrocki:
# [Discover Oasis Filters UI Component](https://app.shortcut.com/oazo-apps/story/6101/tbc-discover-oasis-filters-ui-component)

**Requires https://github.com/OasisDEX/oasis-borrow/pull/1522 to me merged first.**

Proof of concept of working filters. In theory, if everything goes right, it should only require updating endpoint addresses and response object type to make it work with actual data.

image
  
## Changes 👷‍♀️

- Removed dedicated pages for each Discovery page and replaced that with `[slug]` approach from `next.js`,
- created filters for "High-Risk Positions" page,
- created handlers for API that is supposed to load data from data source that will be available somewhere in the future,
- fixed some minor issues with missing `key` props and general html structure in previous Discovery components.
  
## How to test 🧪

- Go to `/discovery/high-risk-positions`,
- change something in the filters,
- check if params from request to mock json file are updating accordingly.


features/discovery/common/DiscoveryWrapperWithIntro.tsx
```diff
@@ -14,7 +14,7 @@ export function DiscoveryWrapperWithIntro({ children }: WithChildren) {
         <Heading variant="header2" sx={{ mb: 2 }}>
           {t('discovery.heading')}
         </Heading>
-        <Text as="p" variant="paragraph1" sx={{ color: 'neutral80', maxWidth: 700, mx: 'auto' }}>
+        <Text variant="paragraph1" sx={{ color: 'neutral80', maxWidth: 700, mx: 'auto' }}>
```
piekczyk, line 17: why semantic was removed?

features/discovery/controllers/DiscoveryControl.tsx
```diff
@@ -0,0 +1,41 @@
+import { DiscoveryFilters } from 'features/discovery/common/DiscoveryFilters'
+import { getDiscoveryData } from 'features/discovery/discoveryApi'
+import { getDefaultSettingsState } from 'features/discovery/helpers'
+import { discoveryPagesMeta } from 'features/discovery/meta'
+import { DiscoveryFiltersSettings, DiscoveryPages } from 'features/discovery/types'
+import React, { useState } from 'react'
+import { Box } from 'theme-ui'
+
+export function DiscoveryControl({ active }: { active: DiscoveryPages }) {
```
piekczyk, line 9: please extract `{ active: DiscoveryPages }` as interface

features/discovery/controllers/DiscoveryControl.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const { endpoint, filters } = discoveryPagesMeta[active]
+  const [settings, setSettings] = useState<DiscoveryFiltersSettings>(
+    getDefaultSettingsState({ filters: filters }),
```
piekczyk, line 12: `getDefaultSettingsState({ filters }),`

features/discovery/meta.tsx
```diff
@@ -1,17 +1,29 @@
...
   iconColor: string
   iconContent: JSX.Element
+  filters: DiscoveryFiltersList
 }
 
-export const DiscoveryPagesMeta: DiscoveryPageMeta[] = [
+export const discoveryPagesMeta: DiscoveryPageMeta[] = [
   {
     kind: DiscoveryPages.HIGH_RISK_POSITIONS,
     id: 'high-risk-positions',
```
piekczyk, line 25: it also looks like good candidate for enum value

piekczyk (changes requested):

features/discovery/common/DiscoveryWrapperWithIntro.tsx
piotrkonowrocki: Because I have to use `withArrow` component inside of it that doesn't have any semantics and renders everything as `div`. Nesting `div` inside of a `p` gives error from react DOM.

I don't feel brave enough to update `withArrow` so it renders as `span` 😅

pages/discovery/index.tsx
```diff
@@ -10,7 +11,7 @@ export async function getServerSideProps(context: GetServerSidePropsContext) {
 
   return {
     redirect: {
-      destination: `/discovery/high-risk-positions${network}`,
+      destination: `/discovery/${discoveryPagesMeta[0].id}${network}`,
```
johnnieskywalker, line 14: maybe `discoveryPagesMeta[0]` could be a variable with name for ex. highRiskPositionsMeta. So one will know which one meta is which, not get confused with order in discoveryPagesMeta array when there will be more pages.

johnnieskywalker (changes requested):

features/discovery/controllers/DiscoveryControl.tsx
```diff
@@ -0,0 +1,41 @@
...
+  )
+  const discoveryData = getDiscoveryData(endpoint, settings)
+
+  // TODO: pass data to table component
+  console.log(discoveryData)
```
johnnieskywalker, line 17: it's useful when testing but could be removed upon merging. 
Besides that component looks and behaves as expected
image

johnnieskywalker (changes requested):

features/discovery/controllers/DiscoveryControl.tsx
johnnieskywalker: requests to mock component are as expected too 
image

features/discovery/controllers/DiscoveryControl.tsx
piotrkonowrocki: Done.

features/discovery/controllers/DiscoveryControl.tsx
piotrkonowrocki: Done.

features/discovery/meta.tsx
piotrkonowrocki: You gave me even better idea, instead of keeping `kind` as `enum` that doesn't have assigned values and just uses auto incrementation, I assigned values from `id` to them and removed `id` completely.

pages/discovery/index.tsx
piotrkonowrocki: Done.

johnnieskywalker (dismissed):

piekczyk (approved):

PR #1546, opened 2022-10-13
Gf/add col updated col ratio

George-field:
# [Added new Col ratio to vault details header](https://app.shortcut.com/oazo-apps/story/5016/update-designs-to-show-current-and-updated-col-ratio)
  
## Changes 👷‍♀️
- Added Col Ratio to vault overview header
  
## How to test 🧪
- Go to manage any vault, you will see a new Collatralization Ratio value next to Current and Next ETH prices

### DIFF
--- a/components/vault/DefaultVaultHeadline.tsx
+++ b/components/vault/DefaultVaultHeadline.tsx
@@ -42,7 +42,7 @@ export function DefaultVaultHeadline({
           subColor: priceChangeColor,
         },
         {
-          label: 'Collateral Ratio',
+          label: t('system.collateral-ratio'),
           value: `${colRatio}%`,
         },
       ]}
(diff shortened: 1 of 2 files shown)


components/vault/DefaultVaultHeadline.tsx
```diff
@@ -39,6 +41,10 @@ export function DefaultVaultHeadline({
           sub: priceChange,
           subColor: priceChangeColor,
         },
+        {
+          label: 'Collateral Ratio',
```
piekczyk, line 45: `label: t('system.collateral-ratio'),`

piekczyk (changes requested):

components/vault/DefaultVaultHeadline.tsx
George-field: ;)

piekczyk (approved):

PR #1567, opened 2022-10-17
Fixed the liquidation price ratio value so the block has appropriate color

marcinciarka:
# Fixed the liquidation price ratio value so the block has appropriate color

## Changes 👷‍♀️
 - fixed the liquidation price ratio values on ManageSectionComponent
  
## How to test 🧪
 - go to your aave position
 - Liquidation price ratio box should have proper background color (pic rel)
image


sennett (approved):
Approved - we should move this logic to the IPosition in the maths lib) in the next iteration.

PR #1601, opened 2022-10-20
Gf/automation unit tests poc

George-field:
I have created a basic PoC to showcase how we could structure tests for pure functions using mocked states to test for correct return values. Ultimately we would want to test the outputs of these functions so that we know what values to expect and be able to handle the results gracefully, as to avoid any issues for a user. 

I ended up removing most tests I had done because I felt that they were not adding much value. 

Imho, To implement a good testing approach it would be best to use tools like jest and enzyme and create a strategy for testing components that interact with these functions as you can test the outputs based on what gets rendered in the UI.


features/automation/optimization/autoTakeProfit/tests/autoTakeProfitTriggerData.test.ts
```diff
@@ -0,0 +1,154 @@
...
+
+describe('autoTakeProfit Form Change tests', () => {
+
+  describe('Given extractAutoTakeProfitData is called', () => {
+
+    it('handles empty input and returns correct value', (done) => {
+
+      const result = extractAutoTakeProfitData([] as any);
+
+      expect(result).to.deep.equal(defaultAutoTakeProfitData)
```
johnnieskywalker, line 18: Nice ide I just wonder maybe in such case we'd like to throw some error or exception that non empty object was expected :thinking:

features/automation/optimization/autoTakeProfit/tests/autoTakeProfitTriggerData.test.ts
```diff
@@ -0,0 +1,154 @@
...
+
+      done()
+    })
+
+    it('handles correct input and returns correct value', (done) => {
+      const data = {
```
johnnieskywalker, line 24: So shouldn't this data be expected at the end of test ?

features/automation/optimization/autoTakeProfit/tests/autoTakeProfitTriggerData.test.ts
```diff
@@ -0,0 +1,154 @@
...
+        maxBaseFeeInGwei: BigNumber,
+        isCloseToCollateral: boolean,
+        replacedTriggerId: number
+      }
+
+      const result = prepareAddAutoTakeProfitTriggerData(data.vaultData as any, data.executionPrice, data.maxBaseFeeInGwei, data.isCloseToCollateral, data.replacedTriggerId)
+
+      expect(result.cdpId).equal(data.vaultData.id)
+      expect(result.kind).equal('addTrigger')
+
```
johnnieskywalker, line 65: good sceanario, we should also check execution price of result IMHO

features/automation/optimization/autoTakeProfit/tests/utils.ts
```diff
@@ -0,0 +1,67 @@
+import BigNumber from "bignumber.js";
+import { Vault } from "blockchain/vaults";
+import { v4 } from 'uuid';
+
+export function generateRandomBigNumber() {
+    return new BigNumber(Math.random())
```
johnnieskywalker, line 6: Could use this one https://mikemcl.github.io/bignumber.js/#random

features/automation/optimization/autoTakeProfit/tests/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export function generateRandomString(type: 'standard' | 'address') {
```
johnnieskywalker, line 9: I don't think that this type parameter for such a function is good practice. There could be just one to generate random address.

features/automation/optimization/autoTakeProfit/tests/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+interface MockVaultOptions {
+    atRiskLevelWarning?: boolean
+    atRiskLevelDanger?: boolean
+    underCollateralized?: boolean
+    atRiskLevelWarningAtNextPrice?: boolean
+    atRiskLevelDangerAtNextPrice?: boolean
+    underCollateralizedAtNextPrice?: boolean
+}
+
+export function createMockVault(options: MockVaultOptions): Vault {
```
johnnieskywalker, line 27: maybe this could be in `vaults.mock.ts`

johnnieskywalker (approved):

PR #1608, opened 2022-10-20
Improve performance when adjusting positon

sennett:
Moves to observable to do gas estimation in a separate async.  Updates the UI with all the values before doing gas estimation.

### DIFF
--- a/features/earn/aave/manage/services/getManageAavePositionStateMachineServices.ts
+++ b/features/earn/aave/manage/services/getManageAavePositionStateMachineServices.ts
@@ -1,7 +1,7 @@
 import { IStrategy, Position } from '@oasisdex/oasis-actions'
 import { BigNumber } from 'bignumber.js'
-import { combineLatest, Observable } from 'rxjs'
-import { first, map } from 'rxjs/operators'
+import { combineLatest, from, Observable, of } from 'rxjs'
+import { first, flatMap, map, startWith, switchMap } from 'rxjs/operators'
 
 import {
   AaveConfigurationData,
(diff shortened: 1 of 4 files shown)


features/earn/aave/manage/state/manageAaveStateMachine.ts
```diff
@@ -339,10 +350,20 @@ export const createManageAaveStateMachine =
...
+              estimatedGasPrice: event.estimatedGasPrice,
+            }
+          } else {
+            // ¯\_(ツ)_/¯
+            return {
+              transactionParameters: event.data?.adjustParams,
+              estimatedGasPrice: event.data?.estimatedGasPrice,
+            }
+          }
+        }),
```
sennett, line 366: If anyone has any ideas about this I'm all ears 👂👂👂

features/earn/aave/manage/state/manageAaveStateMachine.ts
```diff
@@ -82,7 +87,10 @@ export const createManageAaveStateMachine =
         events: {} as ManageAaveEvent,
         services: {} as {
           getParameters: {
-            data: { adjustParams: IStrategy; estimatedGasPrice: HasGasEstimation } | undefined
+            data: {
```
jakubswierczek, line 90: You can't specify `data` in observable. 
Obervables have no data field. They can only send events.

features/earn/aave/manage/services/getManageAavePositionStateMachineServices.ts
```diff
@@ -91,36 +91,59 @@ export function getManageAavePositionStateMachineServices$(
...
-            context.protocolData.position,
-          )
-
-          const gasQty = await txHelpers
-            .estimateGas(callOperationExecutor, {
-              kind: TxMetaKind.operationExecutor,
-              calls: adjustParams.calls as any,
-              operationName: 'CustomOperation',
-              proxyAddress: context.proxyAddress!,
+        ): Observable<{
```
jakubswierczek, line 96: what is the point of that change? 
For me, it's not a stream of values. It's a one-time job. You don't constantly observe that. It's a misunderstanding of observables. 
You only care about the first result. 
I prefer a machine for that because there are 3 steps. 3 atomic operations, not one.

jakubswierczek (changes requested):

features/earn/aave/manage/state/manageAaveStateMachine.ts
sennett: Ah great - maybe this is why I ad to account for the two types above.

features/earn/aave/manage/services/getManageAavePositionStateMachineServices.ts
sennett: Previously this service was a promise that returned simulation data (slow-ish) and gas costs (more slow) inside one promise, that took ~5 seconds to resolve.  The gas cost depends on the simulation data.  My fix was to replace this promise with an Observable that returns the same data in stages.  First the simulation as soon as it's ready, and then the gas cost (and simulation again) when the gas cost becomes ready. 

Observables are both long-lived streams (for example in polling situations) and promises that complete more than once.  We could use a machine here, but in my opinion it would require a lot more code, time and come with a fragility cost, with events, state, actions, spawning, and actors, and more files.  So I went with an Observable instead - in general I think we want to use machines to manage the state powering user journeys, and use other things that have less overhead in other situations.

jakubswierczek: yes, I understand, but if you spawn a machine at the beginning of the parent machine lifecycle, you can send only an update or something like that. 
In the case of Observables or promises are the same. According to the docs, service and actor can be child machine, promise, or observable. 
It doesn't matter what is that. 
Actually in that case Observables don't have `onDone` event:
```
onDone: {
              actions: ['assignTransactionParameters', 'setAdjustOperationType'],
            },
```
Another weak point: you are calling that service only on entry to the `editing` state (and `review`), so if you change the risk, there is no action.
That's why I prefer a machine or callback service if you prefer that https://xstate.js.org/docs/guides/communication.html#invoking-callbacks
But in callbacks, you have to care about subscription and subscription. 

Machines also can send an event after one action. For example, you can return transaction parameters and still execute the gas estimation. 
In your previous implementation example, you have to wait for the gas estimation to see order information. Now, It's faster. But you can easily stop execution if parameters changed. In a machine, you can react to an event and start calculating one more time.

jakubswierczek (approved):

PR #1621, opened 2022-10-24
Marketing the product

marcinciarka:
# [Marketing the product](https://app.shortcut.com/oazo-apps/story/6543/marketing-the-product)

## Changes 👷‍♀️
 - updated the referral banner copy (changed it to more generic "HomepageBanner") + change the logo to stETH + remove the 5% and link this to the opening aave steth position
 - updated the homepage to default to #earn instead of #multiply.
  
## How to test 🧪
 - check the above changes on the home page, should look like this: (note the banner above headline and selected tab on the bottom):
 
 
image

### DIFF
--- a/components/HomePageBanner.tsx
+++ b/components/HomePageBanner.tsx
@@ -5,12 +5,12 @@ import { Flex, Text } from 'theme-ui'
 import { AppLink } from './Links'
 import { Notice } from './Notice'
 
-interface ReferralBannerProps {
+interface HomePageBannerProps {
   heading: string
   link: string
 }
 const handleClose = () => null
-export function ReferralBanner({ heading, link }: ReferralBannerProps) {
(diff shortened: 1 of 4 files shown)


johnnieskywalker (changes requested):
Not sure if this is broken by this PR or something that was merged to dev before, but I can't reproduce this issue on stagging only on this branch. When trying to open landing page with goerli setup, following error is displayed:

image

Seems like stETH yield cannot be calculated
`      response.yield7days && new BigNumber(response.yield7days.yield.netAnnualisedYield),`
in `stEthYield.ts`
please make sure landing page works even if service cannot calculate yield for stETH, probably it cannot on goerli and other testnets, but if it happened on mainnet then please do something else than displaying popup with error on main page

johnnieskywalker (approved):
my bad popup was because of dev helpers
image

PR #1627, opened 2022-10-25
Extract AAVE config

sennett:
# [Extract AAVE config](https://app.shortcut.com/oazo-apps/story/6425/bootstap-skeleton-for-new-strategy)
  
## Changes 👷‍♀️

- Creates /multiply/aave/open/{strategy} route
- Extracts common interface + UI components for aave earn and aave multiply
- Moves AAVE files to non-product-category specific place
  
## How to test 🧪
- Open/manage/close aave positions
- going to `multiply/aave/open/stETHeth` shows placeholders when there is no aave position
- redirection still works when the user already has an aave position


marcinciarka (approved):
Ignoring the smells cause they're all our smells combined 👃

PR #1631, opened 2022-10-25
Skeleton for new strategy

sennett:
# [Skeleton for new strategy](https://app.shortcut.com/oazo-apps/story/6425/bootstap-skeleton-for-new-strategy)

## Changes 👷‍♀️
- Removes simulation machine to unify interface in aave header, so we can inject a different one.
- also starts doing the same to simulate section.
- Consume the open aave context - ISimulation or IPosition inside the simulate component.
- Consume the yields query in the same way as the header, and do the maths based on this.
- Fixes date bug affecting yields stemming from momentjs mutability


features/aave/strategyConfig.ts
```diff
@@ -34,7 +28,7 @@ export const strategyConfig: Record<string, StrategyConfig> = {
       debt: 'ETH',
     },
   },
-  'aave-usdc': {
+  'aave-multiply-steth-usdc': {
```
jakubswierczek, line 31: I thought it was `eth` not `steth`

features/earn/aave/components/AavePositionHeader.tsx
```diff
@@ -37,25 +29,46 @@ export function AavePositionHeader({
...
   }
 
-  const [simulationState] = useActor(simulationActor!)
-  const { context: simulationContext } = simulationState
+  const [minYields, setMinYields] = useState<AaveStEthYieldsResponse | undefined>(undefined)
+  const [maxYields, setMaxYields] = useState<AaveStEthYieldsResponse | undefined>(undefined)
+
+  const { aaveSthEthYieldsQuery } = useAaveContext()
+
+  useEffect(() => {
```
jakubswierczek, line 42: please read it: https://bobbyhadz.com/blog/react-rendered-more-hooks-than-during-previous-render
https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

hooks should be placed before any returns.

jakubswierczek (changes requested):

features/earn/aave/components/AavePositionHeader.tsx
sennett: Good spot.  I will see if I can remove that early return.

features/aave/strategyConfig.ts
sennett: Yes these names are not correct.  I will adjust in another PR.

features/earn/aave/components/AavePositionHeader.tsx
sennett: Resolved this.

jakubswierczek (approved):

PR #1635, opened 2022-10-26
Stop loss with attached xstate events

piekczyk:
# [Stop loss with attached xstate events](https://shortcutLink)

## Changes 👷‍♀️

- created stop loss state machine and attached events to ui changes publish events to monitor how state machine changes over time
  
## How to test 🧪

- run it locally and do action within stop loss form (given example is based on trigger that already exists and user perform updates, testing it in different scenarios wont work because default form state (mapping) have not been implemented)

## State machine
image

- `loadCommonData` and `loadAutomationData` events most likely should be converted to separate state machines (actors) and spawned with their context in `stop loss state machine`. Still I'm not sure how for example `loadAutomationData` would maintain refresh of data when triggers is updated without `send`ing specific event to update its context (as I did in `stop loss machine` with events described above). I tried using services to invoke some asynchronous code but because of that state machine was changing constantly with resulted in warnings in console - `react_devtools_backend.js:4026 Machine given to 'useMachine' has changed between renders. This is not supported and might lead to unexpected results.`.

### DIFF
--- a/features/automation/protection/stopLoss/StopLossContextProvider.tsx
+++ b/features/automation/protection/stopLoss/StopLossContextProvider.tsx
@@ -1,0 +1,33 @@
+import { useInterpret } from '@xstate/react'
+import { StopLossStateMachine } from 'features/automation/protection/stopLoss/state/stopLossStateMachine'
+import { WithChildren } from 'helpers/types'
+import { env } from 'process'
+import React, { createContext, useContext } from 'react'
+
+export const stopLossContext = createContext<StopLossStateMachineContext | undefined>(undefined)
+
+function setupStopLossStateContext({ machine }: { machine: StopLossStateMachine }) {
+  const stateMachine = useInterpret(machine, { devTools: env.NODE_ENV !== 'production' }).start()
+  return {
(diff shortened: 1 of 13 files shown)


features/automation/protection/stopLoss/StopLossContextProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+  children,
+  machine,
+  commonData,
+}: WithChildren & { machine: StopLossStateMachine; commonData: { debt: BigNumber } }) {
+  const stateMachine = useInterpret(machine, { devTools: env.NODE_ENV !== 'production' }).start()
+  const automationContext = useAutomationContext()
+  const [, send] = useActor(stateMachine)
+
+  // PUSH DEPENDENCIES TO MACHINE
+  // MOST LIKELY SHOULD BE CONVERTED TO ACTORS AND SPAWN IN STOP LOSS MACHINE WITH THEIR CONTEXT
```
jakubswierczek, line 38: Nice solution 👍

jakubswierczek (approved):

PR #1646, opened 2022-10-28
🧭 Discover sticky filters

piotrkonowrocki:
# 🧭 Discover sticky filters

Filters follow user while they are scrolling through table.
Works only on desktop, on mobile it was taking too much of a space.
Does not sticky when there are less than two rows in the table.

[Tab-1666960119062.webm](https://user-images.githubusercontent.com/16230404/198587020-c78d0ae6-d9cc-47fc-83e2-c7ca170e5296.webm)

## Changes 👷‍♀️

- Added sticky filters,
- added auto scroll to the top of table after each filtering.
  
## How to test 🧪

- Scroll like a madman 😵

### DIFF
--- a/components/GenericSelect.tsx
+++ b/components/GenericSelect.tsx
@@ -117,6 +117,7 @@ export function GenericSelect({
     menuList: () => ({
       paddingTop: '12px',
       paddingBottom: '12px',
+      maxHeight: '340px',
     }),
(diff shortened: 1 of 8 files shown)


marcinciarka (approved):
:v:

marcinciarka (approved):

PR #1655, opened 2022-11-01
Protocols in 1inch call

jakubswierczek:


helpers/swap.ts
```diff
@@ -64,22 +64,32 @@ export async function oneInchCallMock(
   }
 }
 
+const DEFAULT_PROTOCOLS = ['UNISWAP_V3', 'PMM4', 'UNISWAP_V2', 'SUSHI', 'CURVE', 'PSM']
```
BackToTheCode, line 67: Possibly we add a few more liquidity sources here https://docs.1inch.io/docs/aggregation-protocol/api/swagger

BackToTheCode (approved):

PR #1671, opened 2022-11-09
Gf/move features toggles to db

George-field:
# [Move Feature Flags To Database](https://app.shortcut.com/oazo-apps/story/5331/move-feature-flags-from-local-storage-to-db)

  
## Changes 👷‍♀️
- Added feature flags to schema
- Added a feature flags get endpoint on the api
- update the useFeatureFlag file to accommodate for database feature flags and local-storage

If you look at the code you will notice that in the event that the api request fails when fetching the feature flags, it will fall back to what ever is in local storage. 
  
## How to test 🧪
- update database and run the app with node env in production mode
- set a feature flag in the database, you can use the local storage dev tools to check that is has populated correctly.


piotrkonowrocki (changes requested):
I don't feel like an expert of databases, but there are couple of things that I noticed and they have to be fixed before we can proceed with proper review.

1. Prisma migrations should contain only a diff of changes between current state of database and new one. In here migration seems to contain entire database structure.
2. Structure of `feature_flags` table is counterproductive. In its current state every feature flag is separated column. That means that whenever someone would have to create new feature flag we would need to update database structure and new set of migrations. Instead I strongly suggest that this table should consist of only two columns:
`feature: UNIQUE STRING NOT NULL`
`status: BOOLEAN NOT NULL`
With that setup every feature flag would become an item in database and for checking that one could simply try `SELECT status FROM feature_flags WHERE feature == "autoTakeProfit"`. IF status is returned as `true`, then it's enabled. If status is `false` or there are zero records, feature is disabled.
And thanks to that new feature flags won't require any change to database structure.

helpers/useFeatureToggle.ts
```diff
@@ -46,49 +48,88 @@ const configuredFeatures: Record<Feature, boolean> = {
...
   ShowAaveStETHETHProductCard: true,
-  // your feature here....
+  // your feature here, don't forget to add it to the databse(feature_flags table)....
 }
 
 export function configureLocalStorageForTests(data: { [feature in Feature]?: boolean }) {
   localStorage.setItem(FT_LOCAL_STORAGE_KEY, JSON.stringify(data))
 }
 
+function capitalizeFirstLetter(string: string) {
```
adamskrodzki, line 58: is this name right? unless I'm terribly missing something this function changes first letter `toLowerCase`

helpers/useFeatureToggle.ts
```diff
@@ -46,49 +48,88 @@ const configuredFeatures: Record<Feature, boolean> = {
...
-      localStorage.setItem(FT_LOCAL_STORAGE_KEY, JSON.stringify(merged))
-    }
-  }
+  // fetch toggles from the database
+  axios
+    .get('/api/features')
+    .then((res) => {
+      if (localStorage !== undefined) {
+        // Store values in localstorage becasue if there is a lost connection, features will be able to read from there.
+        if (process.env.NODE_ENV === 'production' && res.data) {
```
adamskrodzki, line 118: is staging `production` ? I think for staging it should also store in database.

helpers/useFeatureToggle.ts
```diff
@@ -46,49 +48,88 @@ const configuredFeatures: Record<Feature, boolean> = {
...
-      ) as ConfiguredFeatures
-      const merged = { ...featuresSourcedFromCode, ...userSelectedFeatures }
-      localStorage.setItem(FT_LOCAL_STORAGE_KEY, JSON.stringify(merged))
-    }
-  }
+  // fetch toggles from the database
+  axios
+    .get('/api/features')
+    .then((res) => {
+      if (localStorage !== undefined) {
```
adamskrodzki, line 116: so if it is undefined database entries are not loaded ?

server/database/migrations/011-feature_flags.sql
```diff
@@ -0,0 +1,233 @@
+-- CreateEnum
```
adamskrodzki, line 1: suspicious ? have you migrated database first ?
why script tries to create all that tables ?

adamskrodzki (changes requested):

server/database/migrations/011-feature_flags.sql
```diff
@@ -0,0 +1,233 @@
...
+    "position_id" TEXT NOT NULL,
+    "collateral_type" TEXT NOT NULL,
+    "collateral_value" DECIMAL(65,30) NOT NULL,
+    "vault_debt" DECIMAL(65,30) NOT NULL,
+    "coll_ratio" DECIMAL(65,30) NOT NULL,
+    "last_action" JSONB NOT NULL
+);
+
+-- CreateTable
+CREATE TABLE "feature_flags" (
```
adamskrodzki, line 161: this table should be key/value 

something in lines of 

`CREATE TABLE "feature_flags" (
"id" SERIAL NOT NULL,
"featureName" varchar(20) NOT NULL,
"status" BOLEAN NOT NULL DEFAULT FALSE
)`

helpers/useFeatureToggle.ts
George-field: Good point, I will check and add that

helpers/useFeatureToggle.ts
George-field: True, missed that, will update

George-field:
> 

Those are both good points @piotrkonowrocki. I also prefer your approach to the table so I'll implement them both.

helpers/useFeatureToggle.ts
George-field: It is production on staging @adamskrodzki

scripts/feature_flags_setup.ts
```diff
@@ -0,0 +1,55 @@
+import { prisma } from '../server/prisma'
+
+// TODD: Will need to get a list of all the feature flags that are currently set
```
adamskrodzki, line 3: resolve TODO

adamskrodzki (approved):
Looks good, just set flags to correct defaults

PR #1686, opened 2022-11-14
created users_who_follow_vaults table

johnnieskywalker:
# [Store users who follow vaults in db](https://app.shortcut.com/oazo-apps/story/5897/store-followed-vaults-in-database-relation-1-address-to-many-cdpids)

  
## Changes 👷‍♀️
- added junction table which specifies which addresses follow which vaults of what type on given chainid
  
## How to test 🧪
- Run dev.sh
- see if migration works
- try to manually add in db records that define user following vault
- Check for edge cases like multiple users can follow the same vault of the same type
- One user can follow multiple vaults
- One user can follow vaults with the same ids on different chains

image

### DIFF
--- a/server/database/migrations/013-add-users-who-follow-vaults.sql
+++ b/server/database/migrations/013-add-users-who-follow-vaults.sql
@@ -7,12 +7,12 @@ CREATE TABLE
     );
 
 -- CreateIndex
-CREATE UNIQUE INDEX "vault_vault_id_chain_id_key" ON "vault" ("vault_id", "chain_id");
+CREATE UNIQUE INDEX "tos_approval_address_chain_id_key" ON "tos_approval" ("address", "chain_id");
+
+-- AddForeignKey
+ALTER TABLE
+    "users_who_follow_vaults" ADD CONSTRAINT "users_who_follow_vaults_user_address_vault_chain_id_fkey" FOREIGN KEY ("user_address", "vault_chain_id") REFERENCES "tos_approval" ("address", "chain_id") ON DELETE RESTRICT ON UPDATE CASCADE;
 
 -- AddForeignKey
 ALTER TABLE
     "users_who_follow_vaults" ADD CONSTRAINT "users_who_follow_vaults_vault_id_vault_chain_id_fkey" FOREIGN KEY ("vault_id", "vault_chain_id") REFERENCES "vault" ("vault_id", "chain_id") ON DELETE RESTRICT ON UPDATE CASCADE;
-
--- AddForeignKey
-ALTER TABLE
-    "users_who_follow_vaults" ADD CONSTRAINT "users_who_follow_vaults_user_address_fkey" FOREIGN KEY ("user_address") REFERENCES "user" ("address") ON DELETE RESTRICT ON UPDATE CASCADE;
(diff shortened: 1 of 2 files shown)


server/database/migrations/013-add-users-who-follow-vaults.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE "users_who_follow_vaults" (
+    "user_address" TEXT NOT NULL,
+    "vault_id" INTEGER NOT NULL,
+    "vault_type" "vault_type" NOT NULL,
```
piotrkonowrocki, line 4: Why do we need `vault_type`, is it anyhow connected to the feature?

server/database/schema.prisma
```diff
@@ -28,12 +28,14 @@ model TosApproval {
 }
 
 model Vault {
-    vault_id      Int       @unique
+    vault_id      Int                    @unique
     type          VaultType
     owner_address String
     chain_id      Int?
+    users         users_who_follow_vaults[]
```
piotrkonowrocki, line 35: Is there any place where we need to display such information? I thought it's going to be based on simple queries from one table.

To see if user follows specific vault something like `SELECT * FROM users_who_follow_vaults WHERE vault_id == vaultId` AND user_address == userAddress`. Or instead of `SELECT` something like `COUNT`, because we only need confirmation if that record exist.

And to get list of all vaults followed by user for overview page `SELECT * FROM users_who_follow_vaults WHERE  user_address == userAddress`.

What am I missing?

server/database/schema.prisma
```diff
@@ -45,13 +47,14 @@ enum VaultType {
...
+    address                    String                 @unique
+    timestamp                  DateTime               @default(now())
     user_that_referred_address String?
-    user_that_referred         User?         @relation("UserToUser", fields: [user_that_referred_address], references: [address])
-    referred_users             User[]        @relation("UserToUser")
+    user_that_referred         User?                  @relation("UserToUser", fields: [user_that_referred_address], references: [address])
+    referred_users             User[]                 @relation("UserToUser")
     weekly_claims              WeeklyClaim[]
     accepted                   Boolean
+    vaults                     users_who_follow_vaults[]
```
piotrkonowrocki, line 57: Same as above - does this table really need to aggregate all followed vaults?

server/database/migrations/013-add-users-who-follow-vaults.sql
johnnieskywalker: We'll need limit of 10 followed vaults per type as defined in here https://app.shortcut.com/oazo-apps/story/5893/validation-limit-number-of-followed-vaults

server/database/schema.prisma
johnnieskywalker: It doesn't it's just definitrion of relation so prisma would know which table is referred by other table, this syntax seem confusing I agree :shrug: 

https://www.prisma.io/docs/concepts/components/prisma-schema/relations/many-to-many-relations#explicit-many-to-many-relations

server/database/migrations/013-add-users-who-follow-vaults.sql
piotrkonowrocki: I spoke with Sam about it, he agreed that we shouldn't care about vault type and users can just hold up to 30 follows.

piotrkonowrocki (approved):
I reviewed and tested, I left one round of comments and concerns and now I think all works well, but I don't feel like database expert so I strongly suggest that it would be good idea if, for example @halaprix could review it as well.

halaprix (approved):
works as intended
image

PR #1703, opened 2022-11-16
🧭 Discover multiselect for "size" and "multiple" filters 

piotrkonowrocki:
# [🧭 Discover multiselect for "size" and "multiple" filters ](https://app.shortcut.com/oazo-apps/story/6797/discover-oasis-multi-select-filters)

image

## Changes 👷‍♀️

- Added different types of filter types as enum: `single`, `multi`, `hidden`,
- set "size" and "multiple" as `multi`,
- set "asset" only for "Most Yield Earned" as `hidden`,
- refined API calls to accept more than one range as `where` filter.
  
## How to test 🧪

- Firstly, you need local Discover database to test it. Since API call were altered, proxy'ing request to staging won't work ,
- after getting yourself localdatabase or connection to staging one (only connection to staging, not API proxy), just test entire filters section.

### DIFF
--- a/features/discover/types.ts
+++ b/features/discover/types.ts
@@ -26,6 +26,12 @@ export enum DiscoverTableVaultStatus {
   TO_STOP_LOSS = 4,
 }
 
+export enum DiscoverFilterType {
+  SINGLE,
+  MULTI,
+  HIDDEN,
+}
+
 export interface DiscoverFiltersSettings {
   [key: string]: string
 }
--- a/features/discover/common/DiscoverFilters.tsx
+++ b/features/discover/common/DiscoverFilters.tsx
@@ -1,7 +1,8 @@
 import { GenericSelect } from 'components/GenericSelect'
 import { DiscoverMultiselect } from 'features/discover/common/DiscoverMultiselect'
-import { DiscoverFiltersList } from 'features/discover/meta'
-import React, { Fragment } from 'react'
+import { DiscoverFiltersList, DiscoverFiltersListItem } from 'features/discover/meta'
+import { DiscoverFilterType } from 'features/discover/types'
+import React from 'react'
 import { Box, Grid } from 'theme-ui'
 
 export function DiscoverFilters({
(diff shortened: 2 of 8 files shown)


features/discover/types.ts
```diff
@@ -26,6 +26,12 @@ export enum DiscoverTableVaultStatus {
   TO_STOP_LOSS = 4,
 }
 
+export enum DiscoverFilterType {
+  SINGLE,
+  MULTI,
+  HIDDEN,
```
johnnieskywalker, line 32: I might not understand the context, but not sure about naming of hidden type. Single and multi are obvious. And does hidden mean there is no filter ?

piotrkonowrocki: I agree, it's a tricky one and it might be confusing. Hidden is a type of a "filter" that sends always the same, fixed params with a request.

Example: on "Most Yield Earned" we don't have "asset" filter, but we don't want to display all types of assets that we have in the database, just vaults with "GUNIV3DAIUSDC1" and "GUNIV3DAIUSDC2" as collateral types. So somehow we need to send them as filters with fixed values without hardcoding them in some kind unholy solution, but using existing flexible connection between filter components and API. So in the configuration of that view "asset" filters looks like that:
```
{
  ...
  type: DiscoverFilterType.HIDDEN,
  options: [
    {
      value: 'GUNIV3DAIUSDC1',
      label: 'GUNIV3DAIUSDC1',
      icon: getToken('GUNIV3DAIUSDC1').iconCircle,
    },
    {
      value: 'GUNIV3DAIUSDC2',
      label: 'GUNIV3DAIUSDC2',
      icon: getToken('GUNIV3DAIUSDC2').iconCircle,
    },
  ],
  ...
}
```
And thanks to that all request are sent with `assets=GUNIV3DAIUSDC1,GUNIV3DAIUSDC2` as a part of querystring. If at any point someone would change type to `MULTI` or `SINGLE` that would result with showing appropriate UI component above the table. But without UI, `HIDDEN` will simply send all of its options with every request.

piekczyk: 200 IQ, I like this approach

piekczyk (approved):

johnnieskywalker (approved):

PR #1721, opened 2022-11-22
Lw/5888 follow unfollow vault

johnnieskywalker:
# [Follow/Unfollow ui component](https://app.shortcut.com/oazo-apps/story/5888/vault-overview-follow-unfollow-vault)
  
## Changes 👷‍♀️
- Created follow button and it's control
  
## How to test 🧪
 - Add `      <FollowButtonControl />` for example in HomepageView.tsx before last </Flex>

- Expect initial state to be Loading
image
- After loading is finished click button, expect loading again, changing state from follow to unfollow
- 
image


johnnieskywalker:
the button atm
image

image

johnnieskywalker:
Added progress stages
image

image

scripts/docker/docker-compose.yaml
```diff
@@ -11,20 +11,20 @@ services:
...
-      - PRIV_KEY_MAINNET=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
-      - TENDERLY_USERNAME=georgi2
-      - TENDERLY_PROJECT=mpoc
-      - USE_DUMMY=${USE_DUMMY}
-      - BLOCK_NUMBER=${BLOCK_NUMBER}
-    ports:
-      - '8545:8545'
-    volumes:
-      - '../.hardhat/cache:/usr/src/mpa/cache'
+  # localnode:
```
piekczyk, line 14: uncomment it

features/homepage/HomepageView.tsx
```diff
@@ -500,6 +501,8 @@ export function Hero({ sx, isConnected }: { sx?: SxStyleProp; isConnected: boole
           }}
         />
       </AppLink>
+
+      <FollowButtonControl />
```
piekczyk, line 505: remove it from here since its not complete implementation

features/follow/common/FollowButtonControl.tsx
```diff
@@ -0,0 +1,27 @@
+import { FollowButton } from 'features/follow/common/FollowButton'
+import { useEffect, useState } from 'react'
+
+export function FollowButtonControl() {
+  const [isFollowing, setIsFollowing] = useState(false)
+  const [isProcessing, setProcessing] = useState(true)
```
piekczyk, line 6: its fine for now but eventually we will need also error handling

piekczyk (changes requested):

piotrkonowrocki (changes requested):
In general, I feel like button is not entirely in line with designs.

- Different sizing and padding inside of a button.
- Different font size.
- Missing border.
- Different hover state, on design background color doesn't change, but border does.
- Loading icon is much smaller than on designs.
- When state of button is changing to loading, there is visible shift in text position. Maybe it would help to keep all types of icons, start and loading, in the same container that is positioned absolute? That way text always stays in the same place and only thing that is changing would be button width, which is totally okay.

scripts/docker/docker-compose.yaml
johnnieskywalker: oh sorry

features/follow/common/FollowButtonControl.tsx
johnnieskywalker: True I'll put TODO mark

piotrkonowrocki (approved):

piekczyk (dismissed):

piotrkonowrocki (approved):

PR #1727, opened 2022-11-23
[DO NOT MERGE] Multicall POC

paszkowskiDamian:

## Changes 👷‍♀️
- item 1
  
## How to test 🧪
- step 1 ...


pages/api/rpc.tsx
```diff
@@ -0,0 +1,93 @@
...
+import * as ethers from 'ethers'
+import { NextApiRequest, NextApiResponse } from 'next'
+
+const RPC = `https://mainnet.infura.io/v3/${process.env.INFURA_PROJECT_ID}`
+
+function getRpcNode(network: string) {
+  switch (network) {
+    case 'mainnet':
+      return `https://mainnet.infura.io/v3/${process.env.INFURA_PROJECT_ID}`
+    default:
```
adamskrodzki, line 12: add goerli as well

pages/api/rpc.tsx
```diff
@@ -0,0 +1,93 @@
...
+      throw new Error('unsupported network')
+  }
+}
+
+function getMultiCallAddress(network: string) {
```
adamskrodzki, line 17: why twice basically same funciton ?

pages/api/rpc.tsx
```diff
@@ -0,0 +1,93 @@
...
+let callsDid = 0
+
+async function makeCall(network: string, calls: any[]) {
+  const response = await axios.post(getRpcNode(network), calls)
+  callsDid += 1
+  return response.data
+}
+export async function rpc(req: NextApiRequest, res: NextApiResponse) {
+  let finalResponse
+  if (process.env.ENABLE_MULTICALL && Array.isArray(req.body) && req.body.every(call => call.method === 'eth_call')) {
```
adamskrodzki, line 57: why not to batch at endpoint instead ?

pages/api/rpc.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const rpcNode = getRpcNode(req.query.network.toString())
+    const provider = new ethers.providers.JsonRpcProvider(rpcNode)
+    const multicall = new ethers.Contract('0x5ba1e12693dc8f9c48aad8770482f4739beed696', abi, provider)
```
adamskrodzki, line 60: looks really bad here :P even if it's same for all networks

adamskrodzki (commented):
Solution seems simple and correct but requires batching calls on UI so might be hard to use it would be better if batching took place inside

adamskrodzki:
fix linter

pages/api/rpc.tsx
paszkowskiDamian: I wanted the multicall to be transparent, and this solution also allowed us to hide infura api key from frontend.

pages/api/rpc.tsx
paszkowskiDamian: it was a poc, It wasn't supposed to be merged at all. I wanted to learn what obstacles we can encounter and I found out. That we cannot query oracle prices like that, and also for some reason not all batches have many calls so solution like this will not decrease the rpc calls this match.

PR #1754, opened 2022-12-01
Feature toggles available outside components

piotrkonowrocki:
# Feature toggles available outside components
  
## Changes 👷‍♀️

- Created `getFeatureToggle` method that allows to use feature toggles outside of components.
  
## How to test 🧪

- Check if hook still works as intended (for feature set to both `false` and `true` as default for all possible states),
- check if `getFeatureToggle` works as intended outside of component context.

### DIFF
--- a/helpers/useFeatureToggle.ts
+++ b/helpers/useFeatureToggle.ts
@@ -89,7 +89,12 @@ export function loadFeatureToggles(testFeaturesFlaggedEnabled: Array<Feature> = 
   }
 }
 
-export function useFeatureToggle(feature: Feature): boolean {
+export function getFeatureToggle(feature: Feature): boolean {
   const userEnabledFeatures = localStorage.getItem(FT_LOCAL_STORAGE_KEY)
+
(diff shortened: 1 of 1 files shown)


helpers/useFeatureToggle.ts
```diff
@@ -89,7 +89,12 @@ export function loadFeatureToggles(testFeaturesFlaggedEnabled: Array<Feature> =
   }
 }
 
-export function useFeatureToggle(feature: Feature): boolean {
+export function getFeatureToggle(feature: Feature): boolean {
```
marcinciarka, line 92: Needs guard
```
typeof localStorage !== 'undefined'
```

marcinciarka (changes requested):

helpers/useFeatureToggle.ts
marcinciarka: Added, thank you.

marcinciarka: Youre welcome.

marcinciarka (approved):

PR #1828, opened 2022-12-19
🧿 User actions for Ajna navigation

piotrkonowrocki:
# 🧿 User actions for Ajna navigation

image
image

#### TODO:
- Add wallet controllers,
- mobile support.
  
## Changes 👷‍♀️

- Created `NavigationOrb` component which is a common component for all user actions and panels in navigation,
- implemented `SwapOrb` and `NotificationsOrb` using `NavigationOrb` so both of them can use the same styling, animation and interaction,
- added "Connect wallet" button to Ajna navigation,
- created `useAccount` hook that contains selected information about user based on global context as a shorthand; for now it's if user is connected and how many positions he has open. That should be further whenever there is opportunity for something to be moved there,
  
## How to test 🧪

- Go to `/ajna` page with `Ajna` feature flag enabled. Without it there should be redirect to homepage,
- test if orbs behavior is similar to those on main landing page (but better 😎),
- test if orbs on main landing page wasn't somehow affected by this changes.

### DIFF
--- a/components/navigation/NavigationMenuOrb.tsx
+++ b/components/navigation/NavigationMenuOrb.tsx
@@ -1,7 +1,7 @@
 import { Icon } from '@makerdao/dai-ui-icons'
 import { useOutsideElementClickHandler } from 'helpers/useOutsideElementClickHandler'
 import { useToggle } from 'helpers/useToggle'
-import React from 'react'
+import React, { PropsWithChildren, useEffect, useRef } from 'react'
 import { Box, Button } from 'theme-ui'
 
 export interface NavigationMenuOrbProps {
@@ -9,6 +9,8 @@ export interface NavigationMenuOrbProps {
   iconSize?: number
   isDisabled?: boolean
   beacon?: boolean | number
+  onClose?: () => void
+  onOpen?: () => void
 }
 
 export function NavigationOrb({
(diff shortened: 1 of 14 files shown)


components/navigation/content/MyPositionsLink.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { Spinner } from 'theme-ui'
+
+export function MyPositionsLink() {
+  const { t } = useTranslation()
+  const { amountOfPositions } = useAccount()
+
+  return (
+    <>
+      {t('my-positions')} (
+      {amountOfPositions || (
```
piekczyk, line 14: I think we shouldn't show 0 when user has 0 positions

piekczyk: image

components/navigation/NavigationMenuOrb.tsx
```diff
@@ -0,0 +1,126 @@
...
+            sx={{
+              position: 'absolute',
+              top: '3px',
+              right: '3px',
+              width: '10px',
+              height: '10px',
+              borderRadius: 'round',
+              bg: 'interactive100',
+              '&::before, &::after': {
+                content: '""',
```
piekczyk, line 65: `content: ''` won't be enough?

piekczyk (changes requested):

components/navigation/content/MyPositionsLink.tsx
piekczyk: on anja infinity loading  
image

piotrkonowrocki: Good catch, fixed.

components/navigation/NavigationMenuOrb.tsx
piotrkonowrocki: No, it won't work.
```
content: '',
```
for `theme.ui` is equivalent of not providing any value to `content` property. `'""`' actually transfers to
```
content: ""
```
in css and only then browser treats `content` as non-empty and renders it.

piekczyk (approved):

PR #1832, opened 2022-12-19
Remove 0 showing next to my positions

piotrkonowrocki:
# Remove `0` showing next to my positions

image
  
## Changes 👷‍♀️

- Fixed issue with condition that was rendering `0` when user has no open vaults next to "My positions" link in navigation.
  
## How to test 🧪

Use two different wallets to test it. One that already has vaults and one that doesn't. When connected, check if number of vaults is displayed correctly and there's no `0` when you don't have any.

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -117,7 +117,7 @@ function PositionsButton({ sx }: { sx?: SxStyleProp }) {
       >
         <Icon name="home" size="auto" width="20" />
       </Button>
-      {amountOfPositions && (
+      {Boolean(amountOfPositions) && (
         <Flex
           sx={{
             position: 'absolute',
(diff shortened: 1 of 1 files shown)


components/Header.tsx
```diff
@@ -264,7 +264,7 @@ function UserDesktopMenu() {
             width="20"
             sx={{ mr: [2, 0, 2], position: 'relative', top: '-1px', flexShrink: 0 }}
           />
-          {t('my-positions')} {amountOfPositions && `(${amountOfPositions})`}
+          {t('my-positions')} {Boolean(amountOfPositions) && `(${amountOfPositions})`}
```
piekczyk, line 267: you could use just `!!` instead of `Boolean`

piekczyk (commented):
there is no longer loading indicator when position number is being loaded, is that expected?

piotrkonowrocki:
> there is no longer loading indicator when position number is being loaded, is that expected?

Yes, loading indicator is something specific that I did for Ajna, it was never a part of main LP. This PR is fixing a small issue that @marcinciarka pointed out - due to a way JSX is working, there was a random zero rendering next to the link in some configurations.

components/Header.tsx
piotrkonowrocki: Fixed.

piekczyk (approved):

PR #1840, opened 2022-12-21
Adjust Aave SL calculations

piekczyk:
# [Adjust Aave SL calculations](https://app.shortcut.com/oazo-apps/story/7310/adjust-aave-sl-calculations-on-ui)

## Changes 👷‍♀️

- adjusted aave sl calculations
- extracted LTV card and used both in aave overview and stop loss 
- removed unused, old details cards
  
## How to test 🧪

- create aave eth/usdc multiply vault using hardhat
- all numbers in stop loss details section and form should be correct (including slider min / max)
- double check whether SL numbers on maker are still fine


components/vault/detailsSection/ContentCardLtv.tsx
```diff
@@ -0,0 +1,54 @@
+import BigNumber from 'bignumber.js'
+import { DetailsSectionContentCard } from 'components/DetailsSectionContentCard'
+import { formatDecimalAsPercent } from 'helpers/formatters/format'
+import { useTranslation } from 'next-i18next'
+import React from 'react'
+
+const getLTVRatioColor = (ratio: BigNumber) => {
+  const critical = new BigNumber(5)
```
piotrkonowrocki, line 8: Would be better if `critical` and `warning` were imported from config.

components/vault/detailsSection/ContentCardLtv.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export function ContentCardLtv({
+  loanToValue,
+  liquidationThreshold,
+  afterLoanToValue,
+}: ContentCardLtvProps) {
+  const { t } = useTranslation()
+
+  return (
+    <DetailsSectionContentCard
```
piotrkonowrocki, line 35: This is very small detail, but I'd like to ask. Why did you change format in how `DetailsSectionContentCard` components are usually set up?

In all cases that I can think of it used to be:
```
const contentCardSettings: ContentCardProps = {
...
}
...

return <DetailsSectionContentCard {...contentCardSettings} />
```
But here you are passing settings directly to props. I'm not saying that approach is worse in any way, just not how other places are formatted.

piotrkonowrocki (changes requested):

piotrkonowrocki (approved):

PR #1890, opened 2023-01-17
🧿 Ajna open borrow form, context and reducto

piotrkonowrocki:
# 🧿 Ajna open borrow form, context and reducto

Very POC version.
About context: it's structure is based on context made for automation and main assumption is that it is going to be reused on both open and manage position, but also at least partially between different products (borrow, earn and multiply). So far some methods or variable names may be misleading for what they are going to be used (global or local context), but until we start working on earn part it's hard to anticipate which parts can be reused or extended and which has to be made dedicated for different product.

Main context ingredients are:
- Environment: external data such as token market prices, users balance etc,
- Position: position state, currently empty as this is just opening flow, so there is no position,
- Form: reducto based state that contains all input from user.

image
  
## Changes 👷‍♀️

- Added two new features to `VaultHeadline` component:
  - outline around token icon (or icons),
  - image label next to title,
- added market prices and balance info handling to ajna,
- created basic context for Ajna,
- created first step of the open borrow form with proper inputs (but they can't be send anywhere so far),
- created empty "Position info view",
- created empty overview for "Setup" view,
- renamed all instances of "debt token" to "quote token" since that's the name they are using at Ajna.
  
## How to test 🧪

- Go to `/ajna/borrow` and select any token pair (best case scenario: test more than one pair),
- test headline and form inputs.


features/ajna/contexts/AjnaProductContext.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  if (!ac) {
+    throw new Error(
+      "AjnaProductContext not available! useAjnaProductContext can't be used serverside",
+    )
+  }
+  return ac
+}
+
+export function AjnaProductContextProvider({
```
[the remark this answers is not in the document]
piekczyk: Also - where are you planning to put things like calculation handlers or calculations itself?

pages/ajna/[product]/[pair].tsx
```diff
@@ -1,29 +1,61 @@
...
+        {([{ collateralBalance }, tokenPriceUSD]) => (
+          <AjnaProductContextProvider
+            collateralBalance={collateralBalance}
+            collateralToken={collateralToken}
+            collateralPrice={tokenPriceUSD[collateralToken]}
+            quoteToken={quoteToken}
+            quotePrice={tokenPriceUSD[quoteToken]}
+          >
+            <WithWalletConnection>
+              <WithTermsOfService>
```
Comment (piekczyk) [inline pages/ajna/[product]/[pair].tsx:51]: I think we will need here TRM (WithWalletAssociatedRisk) as well. I missed it out in the last review.

components/vault/VaultHeadline.tsx
```diff
@@ -31,36 +39,52 @@ export function VaultHeadline({
...
-                mr: tokenData.length - 1 === iconIndex ? 2 : -16,
-              }}
-            />
-          ))}
+        {tokenData instanceof Array && tokenData.length > 0 && (
+          <Box
+            sx={{
+              mr: 2,
+              ...(outline && {
+                filter: `
```
piekczyk, line 62: looks weird but I guess you added it for some reason. Maybe if we catch up to discuss this PR you could provide some explanation :)

features/ajna/borrow/views/AjnaOpenBorrowView.tsx
```diff
@@ -0,0 +1,60 @@
...
+        ]}
+      />
+      <TabBar
+        variant="underline"
+        sections={[
+          {
+            value: 'setup',
+            label: t('setup'),
+            content: (
+              <Grid variant="vaultContainer">
```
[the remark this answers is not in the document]
piotrkonowrocki: Would there be a benefit of that? Without it I have only one wrapper `div`, with `DefaultVaultLayout` there are four 🤔

features/ajna/contexts/AjnaProductContext.tsx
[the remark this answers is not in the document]
piotrkonowrocki: Reducto actions are connected ONLY to the form, not to position state itself. Manage form state is going to be similar, but with couple additional fields (withdraw, repay), so I think that shouldn't be a problem.
Position state and calculations are going to happen outside of a context and Damian is working on specific function for that in library. The idea is that context is going to import that function, pass form state to it and receive simulation and calldata.

Does this answer your question? If not we can try catch on the meet to talk it out.

components/vault/VaultHeadline.tsx
piotrkonowrocki: Added.

pages/ajna/[product]/[pair].tsx
Comment (piotrkonowrocki) [reply pages/ajna/[product]/[pair].tsx:51]: Added.

components/vault/VaultHeadline.tsx
piotrkonowrocki: Yeah, it is weir, but so far I don't know any other way to create outline outside of an image (not images container) than this. Let's talk it out.

piekczyk (approved):

PR #1892, opened 2023-01-17
[SC-7368] Subtract 1 wei from locked amount

jakubswierczek:
# Changes
- We have some isssue with locked collateral amount. When we substact 1 wei it is working 🙉


marcinciarka (approved):
I don't like it, but we need it.

PR #1907, opened 2023-01-19
[SC-7537] position overview many calls

sennett:
# [position overview many calls](https://app.shortcut.com/oazo-apps/story/7537/0-5d-viewing-overview-page-with-a-few-positions-when-disconnected-from-wallet-causes-cascade-of-eth-call-and-eth-getlog-on)

Before:

image
After:

image

  
## Changes 👷‍♀️

- removes `onEveryBlock` from dpm proxy observables
- gets the newly-created DPM proxy from the transaction receipt in the open aave flow instead of getting from top of the app
  
## How to test 🧪

Probably we need a regressing on staging, but this is what we need to test:

- a new DPM proxy is created when needed
- existing empty DPM proxy is consumed
- already-used proxy is not reconsumed
- manage works
    - can adjust risk up and down
    - can withdraw/payback
    - can deposit/borrow
- can close
- DS proxy position still works


blockchain/userDpmProxies.ts
```diff
@@ -27,7 +26,7 @@ export function getUserDpmProxies$(
   const accountGuardGenesisBlock = accountGuardNetworkMap[chainId]
   const accountFactoryGenesisBlock = accountFactoryNetworkMap[chainId]
 
-  return combineLatest(context$, onEveryBlock$).pipe(
+  return combineLatest(context$).pipe(
```
jakubswierczek, line 29: could be just `context$.pipe`, no need to create a new observable. 
But not sure about that. Because I think the RxJS approach was chosen to get changes for each block

jakubswierczek: If we don't want to check it on every block we should return just a promise.

features/stateMachines/dpmAccount/services/getCreateDPMAccountTransactionMachine.ts
```diff
@@ -15,7 +16,11 @@ export function getCreateDPMAccountTransactionMachine(
   context$: Observable<ContextConnected>,
   commonTransactionServices: CommonTransactionServices,
 ) {
-  const service = startTransactionService<CreateDPMAccount>(txHelpers$, context$)
+  const service = startTransactionService<CreateDPMAccount>(
```
jakubswierczek, line 19: `export function startTransactionService<T extends TxMeta, TResult = unknown>`
You can pass the result as a parameter. In that case should be `DPMAccount` or something with the structure:
```
{
vaultId: number, proxy: string, user: string
}
```

features/stateMachines/dpmAccount/state/createDPMAccountStateMachine.ts
```diff
@@ -14,9 +15,13 @@ export interface DMPAccountStateMachineContext {
   refTransactionMachine?: ActorRefFrom<TransactionStateMachine<CreateDPMAccount>>
   error?: string | unknown
   gasData: HasGasEstimation
+  result?: any // transaction result from transactionStateMachine
```
jakubswierczek, line 18: we shouldn't use any.  That's why we have Typescript to avoid that...

features/stateMachines/dpmAccount/services/getCreateDPMAccountTransactionMachine.ts
```diff
@@ -25,3 +30,28 @@ export function getCreateDPMAccountTransactionMachine(
     },
   })
 }
+
+function extractDpmProxyFromTxnReceipt(context: ContextConnected, txnReceipt: any) {
```
jakubswierczek, line 34: please don't use `any`

features/stateMachines/dpmAccount/state/createDPMAccountStateMachine.ts
```diff
@@ -14,9 +15,13 @@ export interface DMPAccountStateMachineContext {
...
 }
 
-export type DMPAccountStateMachineResultEvents = { type: 'DPM_ACCOUNT_CREATED' }
+export type DMPAccountStateMachineResultEvents = {
+  type: 'DPM_ACCOUNT_CREATED'
+  userDpmProxy: UserDpmProxy
```
jakubswierczek, line 23: rename it to `dpmAccount`

jakubswierczek (changes requested):

features/stateMachines/dpmAccount/services/getCreateDPMAccountTransactionMachine.ts
```diff
@@ -25,3 +30,28 @@ export function getCreateDPMAccountTransactionMachine(
...
+  const iface = new ethers.utils.Interface(context.accountFactory.abi.default)
```
marcinciarka, line 35: This could use a better name (`accountFactoryInterface`?)

blockchain/userDpmProxies.ts
sennett: > could be just context$.pipe, no need to create a new observable.

I updated.

> But not sure about that. Because I think the RxJS approach was chosen to get changes for each block

The one regression I found was when creating the proxy, but I will ask Seba who did the original code.  I'm not sure about the everyblock approach in general, because by default we get everything all the time, hence all the RPC calls.  If we only fetch data that we need to render a UI by default, when we need it, then we will be in a better position.  In this case, seeing as we get the DPM data back in the transaction receipt, we can pass that to the machine instead.

> If we don't want to check it on every block we should return just a promise.

We could, but I'm not sure it's feasible at present - unless we convert the promise to an observable before returning, it requires a bigger code change.

jakubswierczek:
After tests:
Test scenario:
- open a DPM in the first tab in your browser (don't close it)
- open the second tab and try to open aave position using created DPM
- return to the first tab and finish opening position.
Result: you have two positions (PositionCreated event) in the same DPM account. 

We can't remove `onEveryBlock` because we need to check if the given proxy wasn't used in the meantime.

blockchain/userDpmProxies.ts
jakubswierczek: please check the test scenario in the last comment.

features/stateMachines/dpmAccount/state/createDPMAccountStateMachine.ts
sennett: Good spot.  Fixed.

features/stateMachines/dpmAccount/services/getCreateDPMAccountTransactionMachine.ts
sennett: Done.

features/stateMachines/dpmAccount/services/getCreateDPMAccountTransactionMachine.ts
sennett: Done.

marcinciarka (approved):

jakubswierczek (approved):

PR #1914, opened 2023-01-20
[SC-7574] Collecting fee in the library (package version bump)

jakubswierczek:
## Changes
- Collecting fee is calculated in the library
- In the library (done in [PR](https://github.com/OasisDEX/oasis-earn-sc/pull/168)) we have a hierarchy in which token we want to get a fee. In that order: `USDC`, `DAI`, `WETH`, `ETH`, `STETH`, `WBTC`. 

How to test:

- Ensure on the FE in its current state multiply fee should always be in USDC, for earn in ETH.


sennett (approved):

J1psy-HUSSLE (commented):
Small hang up in the transition

PR #1917, opened 2023-01-20
added short variant of follow button

johnnieskywalker:
# [Follow button short](https://app.shortcut.com/oazo-apps/story/7595/follow-button-short)

## Changes 👷‍♀️
- added short version of follow button
  
## How to test 🧪
- Add iShort flag like this

```ts
    <FollowButton
      isProcessing={isProcessing}
      isFollowing={isFollowing}
      buttonClickHandler={buttonClickHandler}
      isShort={true}
    />
    ```
- component all works following, unfollowing, just button is shorter   
image

### DIFF
--- a/features/follow/common/FollowButton.tsx
+++ b/features/follow/common/FollowButton.tsx
@@ -8,6 +8,7 @@ interface FollowButtonProps {
   isProcessing: boolean
   isFollowing: boolean
   buttonClickHandler: () => void
+  isShort?: boolean
 }
 
 export function FollowButton(props: FollowButtonProps) {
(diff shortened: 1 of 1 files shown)


piotrkonowrocki (changes requested):
Difference between implementation and design is too big in my opinion. Left one is from design, right is a screen from implementaiton
image
- Icon on design is slightly bigger,
- according to design short version shouldn't have box-shadow, but a border,
- star icon should be a little bit smaller and on implementation is off center.

piotrkonowrocki:
Also, a side note, I feel like in the beginning star was also animated, when its color was changing there was a transition to that. Do I imagine things? 😆

johnnieskywalker:
This changes were included in @piotrkonowrocki 's pull request that is already merged

PR #1918, opened 2023-01-23
Spike for using Xstate component with RXJS or like separated views.

jakubswierczek:
# How to test?
- Go to http://localhost:3000/_example/dpm?network=hardhat
- Try to create a DPM Account. 

# Funcionallity
- We can use CreateDPMAccountView without a machine for the whole flow.
- We can subscribe to events from the machine Proxy View
- We can subscribe to only results. In that case: `UserDpmAccount`

# Problems
- It's not easy to deal with machines with `sendParent`. 
- I created a dummy parent to send all events.
- It can be fixed. I propose solutions in the code.

### DIFF
--- a/pages/_example/dpm/components/containerDPM.tsx
+++ b/pages/_example/dpm/components/containerDPM.tsx
@@ -1,0 +1,60 @@
+import React, { useEffect, useState } from 'react'
+import { map } from 'rxjs/operators'
+import { Box, Container, Grid } from 'theme-ui'
+
+import { DetailsSection } from '../../../../components/DetailsSection'
+import { TabBar } from '../../../../components/TabBar'
+import { CreateDPMAccountView } from '../../../../features/stateMachines/dpmAccount'
+import { useObservable } from '../../../../helpers/observableHook'
+import { useDpmContext } from './dpmProvider'
+export function DPMLog() {
+  const { states$, dpmAccounts$ } = useDpmContext()
+  const onlyEvents = states$.pipe(map((state) => state.event))
+  const [state] = useObservable(onlyEvents)
+  const [dpms] = useObservable(dpmAccounts$)
+  const [content, setContent] = useState<string[]>([])
+  useEffect(() => {
+    setContent([...content, state?.type || ''])
+  }, [state])
+  return (
+    <DetailsSection
+      title={'LOG'}
+      content={
+        <Box>
+          {content.map((c, index) => (
+            <Box key={index}>{c}</Box>
+          ))}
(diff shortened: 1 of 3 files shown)


pages/_example/dpm/components/dpmProvider.tsx
```diff
@@ -0,0 +1,69 @@
...
+const dpmContext = React.createContext<ReturnType<typeof setupDpmContext> | undefined>(undefined)
+
+// This is the easiest solution to deal with `sendParent` in machines.
+// We create a dummy parent machine that will send all events from the child machine.
+// We can subscribe to the events and react to them.
+// Another Solution:
+//  1. Change `sendParent` to send and:.
+//    a. in `getOpenAaveStateMachine.ts` override the sendResultToParent function to `sendParent` instead of `send`
+//    b. don't use `spawn` and pass a service (using for example: useInterpret) to getOpenStateMachine and react to `DPM_ACCOUNT_CREATED` event
+const dummyParent = createMachine({
```
marcinciarka, line 19: This could be imported/exported globally.

sennett: Nice approach.  Do we need to expose this at all?  I think it would be ideal to just have the parent component (the one wrapping `<CreateDPMAccountView />`) not rely on a machine at all.

pages/_example/dpm/components/containerDPM.tsx
```diff
@@ -0,0 +1,60 @@
...
+          {
+            value: 'simulate',
+            label: 'Test',
+            content: (
+              <Grid variant="vaultContainer">
+                <Box>
+                  <DPMLog></DPMLog>
+                </Box>
+                <Box>
+                  <CreateDPMAccountView machine={stateMachine} />
```
sennett, line 50: I think the only thing the rest of the app cares about is getting an empty DPM proxy, or when the flow is abandoned, or an error.  Is there a way of making this interface instead?

```suggestion
                  <CreateDPMAccountView walletAddress={"0x123"} onEmptyDpmProxy={calledWithAnEmptyDpmProxy} onAbandoned={calledWhenUserCancels} onError={calledOnError} />
```

marcinciarka: This would be way better.

jakubswierczek: Yes, it's doable. Just need to create a wrapper. I just wanted to prove that we can use xstate without knowing about it.
I think your proposal is great and can be first steps to mirco frontends or something like that

PR #1932, opened 2023-01-25
aave borrow base page

sennett:
# [Shortcut](https://app.shortcut.com/oazo-apps/story/7645/open-aave-borrow-base-position-page)
  
## Changes 👷‍♀️

- adds base page for Aave borrow
- includes list of allowed collaterals (only eth for now)
- adds borrow strategy config
  
## How to test 🧪
- open, manage, close Earn and Multiply positions to ensure no regression.

### DIFF
--- a/features/aave/services/readPositionCreatedEvents.ts
+++ b/features/aave/services/readPositionCreatedEvents.ts
@@ -20,7 +20,7 @@ type PositionCreatedChainEvent = {
 export type PositionCreated = {
   collateralTokenSymbol: string
   debtTokenSymbol: string
-  positionType: 'Multiply' | 'Earn'
+  positionType: 'Borrow' | 'Multiply' | 'Earn'
   protocol: string
   proxyAddress: string
 }
--- a/features/aave/strategyConfig.ts
+++ b/features/aave/strategyConfig.ts
@@ -1,5 +1,5 @@
 import { ViewPositionSectionComponent } from 'features/earn/aave/components/ViewPositionSectionComponent'
-import { getFeatureToggle } from 'helpers/useFeatureToggle'
+import { Feature, getFeatureToggle } from 'helpers/useFeatureToggle'
 
 import { AaveEarnFaq } from '../content/faqs/aave/earn'
 import { AaveMultiplyFaq } from '../content/faqs/aave/multiply'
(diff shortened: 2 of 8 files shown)


features/aave/services/readPositionCreatedEvents.ts
```diff
@@ -20,7 +20,7 @@ type PositionCreatedChainEvent = {
 export type PositionCreated = {
   collateralTokenSymbol: string
   debtTokenSymbol: string
-  positionType: 'Multiply' | 'Earn'
+  positionType: 'Borrow' | 'Multiply' | 'Earn'
```
marcinciarka, line 23: Why no `ProductType` 🤔

features/aave/strategyConfig.ts
```diff
@@ -58,9 +57,9 @@ export const strategies: Array<IStrategyConfig> = [
     urlSlug: 'ethusdc',
     proxyType: ProxyType.DpmProxy,
     viewComponents: {
-      headerOpen: AaveMultiplyOpenHeader,
-      headerManage: AaveMultiplyManageHeader,
-      headerView: AaveMultiplyManageHeader,
+      headerOpen: AaveOpenHeader,
+      headerManage: AaveManageHeader,
+      headerView: AaveManageHeader,
```
marcinciarka, line 62: Really like those changes.

features/vaultsOverview/pipes/positions.ts
```diff
@@ -297,6 +297,7 @@ export function createAavePosition$(
 }
 
 const mappymap = {
```
marcinciarka, line 299: Not in scope, but we could make `ProductType` as an enum (Borrow: 'Borrow' and so on), ditch the `mappymap` despite the best name ever and use it here.

This can be added to tech debt as well.

pages/borrow/[protocol]/open/[collateralToken].tsx
```diff
@@ -0,0 +1,51 @@
+import { WithConnection } from 'components/connectWallet/ConnectWallet'
```
Comment (marcinciarka) [inline pages/borrow/[protocol]/open/[collateralToken].tsx:1]: I hate that we need to have three basically identical files for this. Need to find a way to reuse (out of scope).

marcinciarka (changes requested):

features/aave/services/readPositionCreatedEvents.ts
sennett: This comes from the event in the smart contracts: https://github.com/OasisDEX/oasis-earn-sc/blob/dev/contracts/actions/common/PositionCreated.sol#L25

pages/borrow/[protocol]/open/[collateralToken].tsx
Comment (sennett) [reply pages/borrow/[protocol]/open/[collateralToken].tsx:1]: Yeah I agree - NextJS was complaining of collisions when using a generic route for all these files, but another way could be to move these files out to somewhere else and have these three files delegate to other files.  Not sure if that is better than the duplication though?

features/vaultsOverview/pipes/positions.ts
sennett: `const mappymap = ProductType` best of both worlds

marcinciarka (approved):

PR #1940, opened 2023-01-26
Changed every instance of stop-color to stopColor

marcinciarka:
# [Fix SVG warnings](https://app.shortcut.com/oazo-apps/story/7225/fix-svg-warnings)

Fixes console warning:
image

I know that files diff shows that everything is broken - it is not.


piotrkonowrocki (changes requested):
I believe we only need an update in `theme/tokenIcons.tsx` file. Normal `svg` files are loaded via html image tag src attribute and aren't a part of jsx interpretation, so it shouldn't matter for next js.

Additionally, using camelcase for attributes in static svg files makes them invalid, so you can't open them outside of next.js enviroment.

PR #1941, opened 2023-01-26
Configure Product Cards to work with AAVE Borrow

BackToTheCode:
# [Configure Product Cards to work with AAVE Borrow](https://app.shortcut.com/oazo-apps/story/7486/open-configure-the-product-cards)
  
## Changes 👷‍♀️
- Add in borrow-against-ETH strategy plus metadata
- Update ProductCardsFilter to work with different types of AAVE strategies
- AAVE Borrow cards now appear on home & borrow pages

Note: Have left /assets out for now as we seem to have ignored these pages for AAVE so far. Can change if we think necessary


components/productCards/ProductCardBorrowAave.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+export function ProductCardBorrowAave({ cardData }: ProductCardBorrowAaveProps) {
+  const { t } = useTranslation()
+  const {
+    aaveSTETHReserveConfigurationData,
+    aaveAvailableLiquidityInUSDC$,
+    getAaveAssetsPrices$,
+  } = useAaveContext()
+  const [aaveReserveState, aaveReserveStateError] = useObservable(aaveSTETHReserveConfigurationData)
```
jakubswierczek, line 32: please make one pipe from it. 
I think we could prepare something like `aaveProductCardData$` which will return only one value. Not `onEveryBlock$`

jakubswierczek (commented):
Make we can first merge https://github.com/OasisDEX/oasis-borrow/pull/1950? It has a bigger priority.

jakubswierczek (dismissed):

jakubswierczek (approved):

PR #1943, opened 2023-01-26
Share button on demand

piotrkonowrocki:
# Share button on demand

Share to twitter button was mistakenly visible not just on existing positions, but also in opening flow.
  
## Changes 👷‍♀️

- Added `shareButton` prop to `vaultHeadline` so it's possible to turn it on and off,
- set `shareButton` to true on all manage positions pages.
  
## How to test 🧪

Visit every type of position (borrow, multiply and earn for maker, multiply and earn for aave and dsr) and see if share button is not visible in open flow, just in manage.

### DIFF
--- a/components/vault/DefaultVaultHeadline.tsx
+++ b/components/vault/DefaultVaultHeadline.tsx
@@ -15,12 +15,14 @@ export function DefaultVaultHeadline({
   priceInfo,
   colRatio,
   followButton,
+  shareButton,
 }: {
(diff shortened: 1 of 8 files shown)


johnnieskywalker:
no share button in opening positions flow

image

image


image

image

button appears at existing positions
image


image

and works too

image


image
image
image
image

johnnieskywalker (approved):

PR #1947, opened 2023-01-27
Component for creating DPM and Allowance flow

marcinciarka:
# Component + hook for creating DPM and Allowance flow

  
## Changes 👷‍♀️
- added new hook - `useFlowState`. The purpose of this hook is to provide access to internals of handling flow states: creating DPM proxy and allowance.
- added new component `FlowSideBar` which can consume `flowState` from the hook above. This provides UI needed for proxy and allowance states.

## How to use it? 🔧 
Typical use case:
```
  const flowState = useFlowState({
    amount: yourDataSource.amount,
    token: yourDataSource.collateral,
    onEverythingReady(params) {
      // we have everything we need, pick the first
      // DPM proxy from "availableProxies" and proceed with your flow
      console.log('onEverythingReady', params)
    },
  })
```
```
  <FlowSidebar {...flowState} />
```
If you don't want to use `onEverythingReady` for whatever reason you can react to the hook changes:
```
{
    dpmMachine, // internals
    allowanceMachine, // internals
    availableProxies, // list of available proxies ⭐
    walletAddress, // just for convenience
    amount, // just for convenience
    token, // just for convenience
    isProxyReady, // self explanatory
    isWalletConnected, // self explanatory
    isAllowanceReady, // self explanatory
    isLoading, // just for the allowance loading state
    isEverythingReady, // self explanatory
  }
```
⭐ internals use first (`[0]`) for allowance and so should you for your flow

For manage flows - everything stays the same, just provide `existingProxy` to the hook and it will be available in `availableProxies`, the hook will ignore getting the list of proxies. This is helpful for the allowance flow (which might need to be updated if user picked minimum allowance for example).
  

  
## How to test 🧪
- use it, tell me what you think and how you feel 💡

### DIFF
--- a/components/FlowSidebar.tsx
+++ b/components/FlowSidebar.tsx
@@ -1,11 +1,6 @@
 import { useActor } from '@xstate/react'
 import { AllowanceView } from 'features/stateMachines/allowance'
 import { CreateDPMAccountViewConsumed } from 'features/stateMachines/dpmAccount/CreateDPMAccountView'
-import {
-  DPMAccountStateMachine,
-  DPMAccountStateMachineEvents,
-} from 'features/stateMachines/dpmAccount/state/createDPMAccountStateMachine'
-import { staticFilesRuntimeUrl } from 'helpers/staticPaths'
 import { useFlowState } from 'helpers/useFlowState'
 import { useTranslation } from 'next-i18next'
 import React from 'react'
--- a/components/FlowSidebar.tsx
+++ b/components/FlowSidebar.tsx
@@ -4,10 +4,11 @@ import {
   DPMAccountStateMachineEvents,
 } from 'features/stateMachines/dpmAccount/state/createDPMAccountStateMachine'
 import { staticFilesRuntimeUrl } from 'helpers/staticPaths'
+import { useFlowState } from 'helpers/useFlowState'
 import { Trans, useTranslation } from 'next-i18next'
 import React from 'react'
 import { Grid, Image, Text } from 'theme-ui'
-import { ActorRefFrom, Sender, StateFrom } from 'xstate'
+import { Sender, StateFrom } from 'xstate'
 
 import { AppLink } from './Links'
 import { ListWithIcon } from './ListWithIcon'
@@ -19,11 +20,9 @@ import {
   VaultChangesInformationItem,
 } from './vault/VaultChangesInformation'
 
-export interface CreateDPMAccountViewProps {
-  machine: ActorRefFrom<DPMAccountStateMachine>
-  isConnected?: boolean
+export type CreateDPMAccountViewProps = {
   noConnectionContent?: JSX.Element
-}
+} & ReturnType<typeof useFlowState>
 
 interface InternalViewsProps {
   state: StateFrom<DPMAccountStateMachine>
@@ -74,7 +73,7 @@ function NoConnectionStateView({
   return <SidebarSection {...sidebarSectionProps} />
 }
 
-function InfoStateView({ state, send }: InternalViewsProps) {
+function DPMInfoStateView({ state, send }: InternalViewsProps) {
   const { t } = useTranslation()
 
   const sidebarSectionProps: SidebarSectionProps = {
@@ -127,7 +126,7 @@ function InfoStateView({ state, send }: InternalViewsProps) {
   return <SidebarSection {...sidebarSectionProps} />
 }
 
-function InProgressView(_: InternalViewsProps) {
+function DPMInProgressView(_: InternalViewsProps) {
   const { t } = useTranslation()
   const sidebarSectionProps: SidebarSectionProps = {
     title: t('dpm.create-flow.proxy-creating-screen.header'),
@@ -163,7 +162,7 @@ function InProgressView(_: InternalViewsProps) {
   return <SidebarSection {...sidebarSectionProps} />
 }
 
-function SuccessStateView({ send }: InternalViewsProps) {
+function DPMSuccessStateView({ send }: InternalViewsProps) {
   const { t } = useTranslation()
 
   const sidebarSectionProps: SidebarSectionProps = {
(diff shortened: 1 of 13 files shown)


helpers/useFlowState.ts
```diff
@@ -0,0 +1,190 @@
...
+    proxyConsumed$,
+    allowanceForAccount$,
+  } = useAppContext()
+  const { context$ } = useAppContext()
+  const { stateMachine: dpmMachine } = setupDpmContext(dpmAccountStateMachine)
+  const { stateMachine: allowanceMachine } = setupAllowanceContext(allowanceStateMachine)
+
+  // wallet connection + DPM proxy machine
+  useEffect(() => {
+    const walletConnectionSubscription = context$.subscribe(({ status, account }) => {
```
jakubswierczek, line 56: `useObservable`

helpers/useFlowState.ts
```diff
@@ -0,0 +1,190 @@
...
+    walletAddress: UserFlowStateReturnType['walletAddress']
+    amount: UserFlowStateReturnType['amount']
+    token: UserFlowStateReturnType['token']
+    isProxyReady: UserFlowStateReturnType['isProxyReady']
+    isWalletConnected: UserFlowStateReturnType['isWalletConnected']
+    isAllowanceReady: UserFlowStateReturnType['isAllowanceReady']
+  }) => void
+}
+
+export function useFlowState({
```
jakubswierczek, line 29: why do we have two responsibilities in one flow? 
we should have one flow for DPMs and one for allowance.

helpers/useFlowState.ts
```diff
@@ -0,0 +1,190 @@
...
+  amount,
+  token,
+  onEverythingReady,
+  existingProxy,
+}: UseFlowStateProps) {
```
jakubswierczek, line 34: I think we should take `owner` as a parameter. Not getting wallet address from `context$`.

helpers/useFlowState.ts
```diff
@@ -0,0 +1,190 @@
...
+import { setupAllowanceContext, setupDpmContext } from './dummyStateMachine'
+import { zero } from './zero'
+
+type UserFlowStateReturnType = ReturnType<typeof useFlowState>
+
+type UseFlowStateProps = {
+  amount?: BigNumber
+  token?: string
+  existingProxy?: string
+  onEverythingReady?: (params: {
```
jakubswierczek, line 18: please make a type for an event

helpers/useFlowState.ts
```diff
@@ -0,0 +1,190 @@
...
+    allowanceMachine,
+    availableProxies,
+    walletAddress,
+    amount,
+    token,
+    isProxyReady: !!availableProxies.length,
+    isWalletConnected,
+    isAllowanceReady,
+    isLoading, // just for the allowance loading state
+    isEverythingReady: isAllowanceReady, // just for convenience, allowance is always the last step
```
jakubswierczek, line 188: What about the situation when the user has approval for a given amount?

piotrkonowrocki (changes requested):
Few comments about allowance stage:
- Before I select any option, the button should be disabled.
- Loader never stops spinning. Also, there's a funny effect, every time when I hover over anything inside of a sidebar it resets and try spinning from the top 💃
image

After selecting amount and confirming transaction, the message in sidebar is incomplete. Shouldn't it say "(...) Your Smart Defi Account will be able to interact with up to: **AMOUNT**", since we know what was selected?
image
image

Shouldn't there be something else on allowance success screen? An image similar to the on on dpm success screen?
image

When I click continue on success screen, after a second or so success screen is shown up again. Data from `onEverthingReady` seems to be correct:
image

jakubswierczek (changes requested):

components/FlowSidebar.tsx
```diff
@@ -0,0 +1,245 @@
...
+  const allowanceConsidered = token !== 'ETH' && amount
+
+  if (!isWalletConnected) {
+    return <NoConnectionStateView noConnectionContent={noConnectionContent} />
+  }
+  if (!isProxyReady && !isAllowanceReady) {
+    switch (true) {
+      case dpmState.matches('idle'):
+      case dpmState.matches('txFailure'):
+        return <DPMInfoStateView state={dpmState} send={dpmSend} />
```
piotrkonowrocki, line 226: Allowance is nicely split and separated into `AllowanceView`, I suggest moving DPM related components from this file to something similar.

helpers/useFlowState.ts
piotrkonowrocki: This is expected behavior. For our current needs there isn't a situation where you need only DPM or only allowance flows - they always come as a pair.

We also spoke that in the future that could be refined into a wrapped that merges two separated flow, but for now this is exactly what we need.

helpers/useFlowState.ts
```diff
@@ -0,0 +1,190 @@
...
+    isProxyReady: UserFlowStateReturnType['isProxyReady']
+    isWalletConnected: UserFlowStateReturnType['isWalletConnected']
+    isAllowanceReady: UserFlowStateReturnType['isAllowanceReady']
+  }) => void
+}
+
+export function useFlowState({
+  amount,
+  token,
+  onEverythingReady,
```
piotrkonowrocki, line 32: We are going to need one more callback: `onBack`. If function is passed, there should be a `textButton` passed to sidebar that would trigger this callback. Thanks to that you can always get back from allowance or dpm step back to from wherever you came from.

And I guess that button should be visible only on first steps of both flows. If we put it on progress, success or error screen then it doesn't make any sense.

piotrkonowrocki (changes requested):

helpers/useFlowState.ts
marcinciarka: We're never gonna check the DPM accounts of another address than wallet and this simplifies things. If you have a strong opinion please reopen :v:

helpers/dummyStateMachine.ts
```diff
@@ -0,0 +1,59 @@
...
+      }
+    })
+
+  const service = useInterpret(machine, { parent: parentService }).start()
+
+  const states$ = from(service)
+  return {
+    stateMachine: service,
+    states$,
+    dpmAccounts$: createdDpmAccount.asObservable(),
```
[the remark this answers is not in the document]
marcinciarka: Nope in both cases, removed.

helpers/useFlowState.ts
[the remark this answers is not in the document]
marcinciarka: This is ultimately _always_ the last step, thats why I named it like this.

helpers/useFlowState.ts
marcinciarka: There's a check for that in `allowanceSubscription`.

components/FlowSidebar.tsx
marcinciarka: Done.

helpers/useFlowState.ts
marcinciarka: Done, `onGoBack`.

jakubswierczek (dismissed):

helpers/useFlowState.ts
piotrkonowrocki: ❤️

piotrkonowrocki (approved):

PR #1957, opened 2023-02-01
🧿 Ajna header selector

piotrkonowrocki:
# [🧿 Ajna header selector](https://app.shortcut.com/oazo-apps/story/7327/select-collateral-token-drop-down)

~~For now, just the selector part without table and connection to anything.~~
Now also includes table with fake data randomized, but based on curated tokens list. For now missing parts are connection to the blockchain ~~and marking some token pairs as not yet available~~.

image

Additionally, `HeaderSelector` can be used in any context as it inherits font styles from its parent.

image
  
## Changes 👷‍♀️

- Created `HeaderSelector` component,
- created `AjnaSelectorController` which could be reused for earn and multiply,
- added support for JSX to discover table - that should be done when I was working on user's position, but I was fixating myself on keeping that table pure data. Not anymore.
  
## How to test 🧪

- Go to `/ajna/open/borrow` and see how selector behaves. ~~Changing its value won't affect anything on that page (for now),~~
- ignore 4 Security Hotspots from sonarcloud - `Math.random` in `useEffect` is just temporary for dummy data.


features/ajna/common/consts.ts
```diff
@@ -1,14 +1,31 @@
...
+    ETH: ['USDC'],
+    RETH: ['ETH'],
+    USDC: ['ETH', 'WBTC'],
+    WBTC: ['USDC'],
+    WSTETH: ['USDC', 'ETH'],
+  },
+  multiply: {},
+  earn: {},
+}
+export const ajnaUnavailablePairs: AjnaPairs = {
```
johnnieskywalker, line 19: I'd suggest to name it commingSoonPairs, well AFAIK we won't support all possible pairs but some recommended list, `unavailable` ones are in fact ones that in UI are `Comming Soon`

components/HeaderSelector.tsx
```diff
@@ -0,0 +1,166 @@
...
+              onClick={() => {
+                setIsOpen(false)
+                setSelected(option)
+                if (onChange) onChange(option)
+              }}
+            >
+              {option.icon && (
+                <Icon size={32} sx={{ flexShrink: 0, mr: '12px' }} name={option.icon} />
+              )}
+              {option.label}
```
johnnieskywalker, line 159: Even this hypothetical edge case is fine :ok_hand: 
image

features/discover/common/DiscoverTableHeading.tsx
```diff
@@ -9,7 +9,7 @@ export function DiscoverTableHeading({ children }: PropsWithChildren<{}>) {
       sx={{
         position: 'relative',
         px: ['24px', null, null, 4],
-        pt: [3, null, null, 4],
+        mt: [3, null, null, 4],
```
johnnieskywalker, line 12: I just quickly checked if everything remains great on discovery table it does :heavy_check_mark: 

https://user-images.githubusercontent.com/6871923/216403501-324e6839-faa1-424d-a9db-c91ea55d4e9f.mp4

johnnieskywalker (changes requested):

features/ajna/common/consts.ts
piotrkonowrocki: Fixed.

components/HeaderSelector.tsx
piotrkonowrocki: Well, in that case it doesn't look very good imo 😅 But guni tokens can't be used as a collateral and there's no way that amount of text could fit in there. If that situation ever happen, font would have to be smaller for mobiles. But it's not the case right now.

piekczyk (approved):

piekczyk (approved):

johnnieskywalker (approved):

PR #1967, opened 2023-02-03
Ajna actions handling boilerplate

piekczyk:
# [Ajna actions handling boilerplate](https://shortcutLink)

## Changes 👷‍♀️

- tbu
  
## How to test 🧪

- tbu


blockchain/config.ts
```diff
@@ -484,6 +500,12 @@ const goerli: NetworkConfig = {
     aaveV3PoolDataProvider,
     goerliAddresses.AAVE_V3_POOL_DATA_PROVIDER,
   ),
+  ajnaPollInfo: contractDesc(ajnaPoolInfo, '0x32DA971821E6d88FCcD9861EbaCF8590553A2aA8'),
+  ajnaProxyActions: contractDesc(ajnaProxyActions, '0x2010D2d932b467928313F86653b28E22A9d6889b'),
+  ajnaPollPairs: {
+    'WBTC-USDC': contractDesc(ajnaPool, '0xcCbD19488f4e63319c2f11e156ccfA26Ce99c657'),
```
piotrkonowrocki, line 506: To discuss: how to marry this thing with `ajnaPairs` in `oasis-borrow/features/ajna/common/consts.ts`?

features/ajna/borrow/sidebars/AjnaBorrowFormContent.tsx
```diff
@@ -15,9 +16,14 @@ export function AjnaBorrowFormContent() {
...
   const {
     environment: { collateralToken, flow, product, quoteToken },
+    form: {
+      state: { action },
+    },
     steps: { currentStep, isStepValid, isStepWithBack, setNextStep, setPrevStep },
   } = useAjnaBorrowContext()
 
+  // TODO use here proxyAddress from DPM state machine
+  const txHandler = useAjnaTxHandler({ proxyAddress: '0xF5C0D205a00A5F799E3CFC4AC2E71C326Dd12b76' })
```
piotrkonowrocki, line 36: I suggest doing it the semi-right way alrady.

You could add `dpmProxyAddress` to form reducto, hardcode this as a default value, but in here get it from `form: { state: { dpmProxyAddress } }`.

features/ajna/borrow/sidebars/AjnaBorrowFormContent.tsx
```diff
@@ -70,7 +76,14 @@ export function AjnaBorrowFormContent() {
         ? {
             url: '/connect',
           }
-        : { action: setNextStep }),
+        : {
+            action: async () => {
```
piotrkonowrocki, line 80: We have to think about some external method for that, because it's going to grow into uncontrolled monster and there's too much logic in the UI. That also relates to my last PR.

But for now, I guess we need
```
if (currentStep === 'transaction' && action) txHandler()
else setNextStep()
```
so it won't throw error on transaction step every time when you click button.

features/ajna/borrow/useAjnaTxHandler.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  const quotePrecision = getToken(quoteToken).precision
+  const collateralPrecision = getToken(collateralToken).precision
+  const defaultPromise: Promise<ActionData> = new Promise((resolve) => resolve({} as ActionData))
+
+  const dependencies = {
+    provider: rpcProvider,
+    poolInfoAddress: context.ajnaPollInfo.address,
+    ajnaProxyActions: context.ajnaProxyActions.address,
+    WETH: context.tokens.ETH.address,
```
piotrkonowrocki, line 59: This thing is still needed?

features/ajna/borrow/useAjnaTxHandler.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  switch (formState.action) {
+    case 'open':
+      if (!depositAmount || !generateAmount) {
+        return defaultPromise
+      }
+      return await strategies.ajna.open(
```
piotrkonowrocki, line 67: @paszkowskiDamian, shouldn't there be `strategies.ajna.**borrow**.open`? I mean on the library part, Seba is just importing that as it is.

features/ajna/borrow/useAjnaTxHandler.ts
```diff
@@ -0,0 +1,150 @@
...
+        {
+          poolAddress: context.ajnaPollPairs[tokenPair].address,
+          collateralAmount: depositAmount.times(ten.pow(collateralPrecision)),
+          debtAmount: generateAmount.times(ten.pow(quotePrecision)),
+          dpmProxyAddress: proxyAddress,
+          // TODO hardcoded for now, but will be moved eventually to library
+          price: new BigNumber(16821273),
```
piotrkonowrocki, line 74: What is this price?

features/ajna/borrow/useAjnaTxHandler.ts
```diff
@@ -0,0 +1,150 @@
...
+  const { txHelpers$, context$ } = useAppContext()
+  const [txHelpers] = useObservable(txHelpers$)
+  const [context] = useObservable(context$)
+  const {
+    form: { state },
+    tx: { setTxDetails, setSimulation },
+    environment: { collateralToken, quoteToken, ethPrice },
+  } = useAjnaBorrowContext()
+
+  const [txData, setTxData] = useState<TxData | null>(null)
```
piotrkonowrocki, line 111: Is there a difference in logic for you if you use `null` or `undefined`? Because you could do something like that:
```
useState<TxData>()
```
and without passing nulls, the type for this is `txData | undefined`.

features/ajna/common/types.ts
```diff
@@ -18,3 +18,5 @@ export type AjnaPairs = {
     [key: string]: string[]
   }
 }
+
+export type AjnaPoolPairs = 'WETH-USDC' | 'WBTC-USDC'
```
piotrkonowrocki, line 22: Is there a way that instead of type guard we can use `AjnaPairs` for validation?

features/ajna/contexts/AjnaProductContext.tsx
```diff
@@ -69,6 +72,7 @@ interface AjnaBorrowContext {
   position: AjnaBorrowPosition
   steps: AjnaBorrowSteps
   tx: AjnaBorrowTx
+  simulation?: AjnaBorrowPosition
```
piotrkonowrocki, line 75: There is `position: AjnaBorrowPosition` already above. Also, I'm not sure if making that optional is the best solution, maybe some default state with zeros would work better here?

piotrkonowrocki (changes requested):

features/ajna/contexts/AjnaProductContext.tsx
piekczyk: I did it in that way because we will need separate place to keep existing position data (for manage view) and place to keep and update simulations based on user input. The interface will be the same.

features/ajna/borrow/useAjnaTxHandler.ts
piekczyk: it is the lowest price you can get to when depositing specific amount of collateral (it is slippage-like parameter)

features/ajna/borrow/useAjnaTxHandler.ts
piekczyk: yes as library is not aware of network that is used on FE

features/ajna/borrow/sidebars/AjnaBorrowFormContent.tsx
piotrkonowrocki: Even better: `isStepWithTransaction`.

blockchain/config.ts
piekczyk: as discussed we will use `export type AjnaPoolPairs = keyof Context['ajnaPollPairs']`

features/ajna/common/types.ts
piekczyk: as discussed we will use `export type AjnaPoolPairs = keyof Context['ajnaPollPairs']`

features/ajna/borrow/state/ajnaBorrowFormReducto.ts
```diff
@@ -34,6 +35,10 @@ interface AjnaBorrowFormActionsUpdatePayback {
   paybackAmount?: BigNumber
   paybackAmountUSD?: BigNumber
 }
+interface AjnaBorrowFormActionsUpdateProxyAddress {
```
piotrkonowrocki, line 38: How about using the power of Reducto 🪄, scratch that part and just use
```
updateState('proxyAddress', ...)
```
?

piotrkonowrocki (changes requested):

features/ajna/borrow/state/ajnaBorrowFormReducto.ts
piekczyk: yep, updated

piotrkonowrocki (approved):

PR #1980, opened 2023-02-07
Ajna inputs debounce

piekczyk:
# [Ajna inputs debounce](https://shortcutLink)

## Changes 👷‍♀️

- wrapped ajna tx handlers in `useDebounceEffect` to avoid potential spamming of rpc calls  
- added `isLoading` property in ajna reducto

## How to test 🧪

- when user inputs something in ajna form `getTxDetails` method should be called after delay of 250 ms


features/ajna/borrow/sidebars/AjnaBorrowFormContent.tsx
```diff
@@ -65,8 +67,8 @@ export function AjnaBorrowFormContent({
     ),
     primaryButton: {
       label: t(getPrimaryButtonLabelKey({ currentStep, product, dpmAddress, walletAddress })),
-      disabled: !!walletAddress && (!isStepValid || isAllowanceLoading),
-      isLoading: !!walletAddress && isAllowanceLoading,
+      disabled: !!walletAddress && (!isStepValid || isAllowanceLoading || isLoadingSimulation),
+      isLoading: (!!walletAddress && isAllowanceLoading) || isLoadingSimulation,
```
piotrkonowrocki, line 71: Correct me if I'm wrong, but I think it should be
```suggestion
      isLoading: !!walletAddress && (isAllowanceLoading || isLoadingSimulation),
```
Unless wallet is connected we don't need to display loading, because that button leads only to connection page, not to next step.

features/ajna/contexts/AjnaProductContext.tsx
```diff
@@ -157,9 +167,16 @@ export function AjnaBorrowContextProvider({
       form: { ...prev.form, state: form.state },
       steps: setupStepManager(),
       tx: setupTxManager(),
-      simulation,
     }))
-  }, [props.collateralBalance, props.quoteBalance, form.state, currentStep, txDetails, simulation])
+  }, [
+    props.collateralBalance,
+    props.quoteBalance,
+    form.state,
```
piotrkonowrocki, line 174: Shouldn't that be also updated in the same way how you split dependency array in `features/ajna/borrow/useAjnaTxHandler.ts`?

piotrkonowrocki (changes requested):

features/ajna/contexts/AjnaProductContext.tsx
piekczyk: no, we need to pass here the whole form state in order to refresh context state with most recent form state

PR #1983, opened 2023-02-08
🧿 Ajna position loading

piotrkonowrocki:
# 🧿 Ajna position loading

So far without displaying its data on the UI.
  
## Changes 👷‍♀️

- Created observable that loads position metadata based on it's ID from url,
- utilized method from library for loading position data,
- added handling for owner and non owners view in manage position - banner and hiding primary button (`isOwner` is set to `true` if user is on open flow since there is no real owner of that page),
- added redirect to `404` when there is no Ajna position of given ID,
  
## How to test 🧪

- Go to `/ajna/position/1?network=goerli`,
- see if page is loading and test its interface from three context: owner, non owner and disconnected wallet.


helpers/getPositionIdentity.ts
```diff
@@ -0,0 +1,18 @@
+import { getAddress } from 'ethers/lib/utils'
+
+function safeGetAddress(address: string) {
+  try {
+    return getAddress(address)
+  } catch (e) {
+    return undefined
```
piekczyk, line 7: maybe it is worth to add here `console.warn`, in that case it will be easier to debug

features/ajna/controls/AjnaProductController.tsx
```diff
@@ -82,6 +84,34 @@ export function AjnaProductController({
     ),
   )
 
+  useEffect(() => {
```
piekczyk, line 87: this `useEffect` looks like something that could be combined with `dpmPositionData` within a single observable and managed on the context level, let me know what do you think

piekczyk (changes requested):

helpers/getPositionIdentity.ts
piotrkonowrocki: I'm not sure if that's the best way. Look when it's used, `getPositionIdentity` tries to return either `vaultId` or `walletAddress`. It never returns both and in most cases we are going to use `vaultId`. That means that we are going to get that warning on each refresh of positions page.

features/ajna/controls/AjnaProductController.tsx
piotrkonowrocki: Fixed. The `useEffect` has to stay because it's still needed to update details about product type, collateral and quote tokens, but I removed library method usage from controlled and moved it into single observable.

Note: There are still two separated observables: `ajnaPosition$` and `dpmPositionData$`. The second one is universal and could be used as well to load aave position or any other that utilizes DPM.

piekczyk (approved):

PR #1988, opened 2023-02-08
🧿 Ajna order informations and after pills

piekczyk:
# [Ajna order informations and after pills](https://app.shortcut.com/oazo-apps/story/7924/ajna-minor-features)

## Changes 👷‍♀️

- added order information based on simulation (not all are defined since library is not ready yet)
- added after pills
  
## How to test 🧪

- use ajna feature toggle
- order information in ajna form and after pills should display data from simulation


features/ajna/borrow/overview/ContentCardCollateralLocked.tsx
```diff
@@ -26,16 +26,21 @@ export function ContentCardCollateralLocked({
...
 
   const contentCardSettings: ContentCardProps = {
     title: t('ajna.borrow.common.overview.collateral-locked'),
-    value: formatted.loanToValue,
+    value: formatted.collateralLocked,
     unit: collateralToken,
-    footnote: formatted.collateralLockedUSD,
+  }
+
+  if (!collateralLocked.isZero()) {
```
piotrkonowrocki, line 42: Why? If there is no collateral we can just display $0 as it's value instead of hiding it.

features/ajna/borrow/overview/ContentCardPositionDebt.tsx
```diff
@@ -26,21 +26,25 @@ export function ContentCardPositionDebt({
...
 
   const contentCardSettings: ContentCardProps = {
     title: t('ajna.borrow.common.overview.position-debt'),
-    value: formatted.loanToValue,
+    value: formatted.positionDebt,
     unit: quoteToken,
-    footnote: formatted.collateralLockedUSD,
+  }
+
+  if (!positionDebt.isZero()) {
```
piotrkonowrocki, line 41: Same question as in `features/ajna/borrow/overview/ContentCardCollateralLocked.tsx`.

features/ajna/borrow/sidebars/AjnaBorrowFormContent.tsx
```diff
@@ -66,15 +67,35 @@ export function AjnaBorrowFormContent({
...
+        !!walletAddress &&
+        (isAllowanceLoading || isSimulationLoading || isTxInProgress || isTxWaitingForApproval),
       ...(!walletAddress && currentStep === editingStep
         ? {
             url: '/connect',
           }
         : {
             action: async () => {
+              if (isTxSuccess && id) {
+                push(`/ajna/position/${id}`)
```
piotrkonowrocki, line 97: Please use `url` prop in `primaryButton` instead of action and manually pushing via next.js hook. If you pass `url` prop, entire button is changed into proper link that keeps `network` in querystring.

features/ajna/borrow/sidebars/AjnaBorrowFormContentDeposit.tsx
```diff
@@ -22,7 +23,7 @@ export function AjnaBorrowFormContentDeposit() {
       {depositAmount?.gt(0) && (
         <>
           <SidebarResetButton clear={() => dispatch({ type: 'reset' })} />
-          <AjnaBorrowFormOrder />
+          {simulation && <AjnaBorrowFormOrder />}
```
piotrkonowrocki, line 26: Won't it make a jumping effect in the form? First, after user inputs something into input he's going to get reset, then only after simulation is loaded order information? Maybe when simulation isn't available we should display all values inside of it as a skeleton loaders? But only on initial loading, not every time when values are getting updated.

features/ajna/borrow/sidebars/AjnaBorrowFormOrder.tsx
```diff
@@ -1,32 +1,69 @@
...
 
 export function AjnaBorrowFormOrder() {
   const { t } = useTranslation()
 
+  const {
+    environment: { collateralToken, quoteToken },
+    position: { currentPosition, simulation },
+  } = useAjnaBorrowContext()
+
+  const collateralLocked = currentPosition?.collateralAmount || zero
```
piotrkonowrocki, line 17: Same comment as in `features/ajna/borrow/overview/AjnaBorrowOverviewWrapper.tsx`. We won't need to pass `zero` for safety here, I suggest skipping those values for now.

piotrkonowrocki (changes requested):

features/ajna/borrow/overview/AjnaBorrowOverviewWrapper.tsx
```diff
@@ -50,8 +62,10 @@ export function AjnaBorrowOverviewWrapper() {
               collateralToken={collateralToken}
               quoteToken={quoteToken}
               cost={zero}
```
piotrkonowrocki, line 64: Don't we have access to `cost`? I think Damian told that it should be available as one of the props in interface.

piotrkonowrocki (changes requested):

piotrkonowrocki (approved):

PR #1992, opened 2023-02-09
🧿 Refresh pool selector

piotrkonowrocki:
🧿 Refresh pool selector

On initial load pool selector wasn't showing pools that are available in context config, just items that are coming soon.
  
## Changes 👷‍♀️

- Fixed issue above.
  
## How to test 🧪

Go to page with pool selector and see if pair ETH/USDC is available to select from the beginning.

### DIFF
--- a/features/ajna/controls/AjnaSelectorController.tsx
+++ b/features/ajna/controls/AjnaSelectorController.tsx
@@ -33,23 +33,31 @@ export function AjnaSelectorController({ product }: AjnaSelectorControllerProps)
   const { context$ } = useAppContext()
   const [context] = useObservable(context$)
   const [hash] = useHash()
-  const defaultOptionValue = hash.length ? hash.replace('#', '') : DEFAULT_SELECTED_TOKEN
   const ref = useRef<HTMLDivElement>(null)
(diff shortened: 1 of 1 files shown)


features/ajna/controls/AjnaSelectorController.tsx
```diff
@@ -33,24 +33,32 @@ export function AjnaSelectorController({ product }: AjnaSelectorControllerProps)
...
-      (pool) => pool.split('-')[0],
-    ),
-  )
-    .sort()
-    .map((token) => ({
-      label: token,
-      value: token,
-      icon: getToken(token).iconCircle,
-    }))
+  const [options, setOptions] = useState<HeaderSelectorOption[]>(getOptions())
```
piekczyk, line 37: instead of using `useState` with `useEffect` I think you could just use `useMemo` to define `options` with dependency array set to `context?.ajnaPoolPairs`

piotrkonowrocki: Now that you mention it, that totally makes sense. Fixed.

piekczyk (approved):

PR #1999, opened 2023-02-13
 🧿 Ajna borrow redirect to new position

piekczyk:
# [Ajna borrow redirect to new position](https://app.shortcut.com/oazo-apps/story/7924/ajna-minor-features)

## Changes 👷‍♀️

- added observable for fetching position id based on user dpm proxy

  
## How to test 🧪

- `Go to position` button should redirect user to newly created position (borrow)


blockchain/userDpmProxies.ts
```diff
@@ -147,3 +147,34 @@ export function getUserDpmProxy$(
     }),
   )
 }
+
+export function getPositionIdFromDpmProxy$(
+  context$: Observable<Context>,
+  dpmProxy?: string,
+): Observable<string | undefined> {
+  if (!dpmProxy) {
```
piotrkonowrocki, line 155: Why not just set that parameter as `required` instead of checking if it exist?

features/ajna/contexts/AjnaProductContext.tsx
```diff
@@ -107,6 +109,16 @@ export function AjnaBorrowContextProvider({
   if (!isAppContextAvailable()) return null
 
   const form = useAjnaBorrowFormReducto({})
+  const { positionIdFromDpmProxy$ } = useAppContext()
```
piotrkonowrocki, line 112: What do you think about changing that into a way where you don't use observable to constantly check for ID of that position, but instead of that you would simply subscribe just once to it after transaction is done? Instead of `useObservable` simple `positionIdFromDpmProxy$.subscribe` in transaction handler after transaction is finished with loader still left on `true` until ID is known.

piotrkonowrocki (changes requested):

features/ajna/contexts/AjnaProductContext.tsx
piekczyk: added `first()` to that observable so position id is read only one time per dpm proxy address change

piotrkonowrocki (approved):

PR #2019, opened 2023-02-14
🧿 Partial CBETH support

piotrkonowrocki:
# 🧿 Partial CBETH support

Missing data that would allow to display this token as a product card, but has enough to be displayed in pool selector table.
  
## Changes 👷‍♀️

- Added partial metadata for CBETH,
- added CBETH to Ajna pool selector as a coming soon.
  
## How to test 🧪

See if CBETH is available as a collateral token in dropdown on `/ajna/borrow`.


johnnieskywalker (approved):
CBETH is available in ajna/borrow section as expected
image

It's marked as coming soon as expected

Everything works in mobile, also as coming soon with details missing that are going to be added later as specified in pull request

image

PR #2020, opened 2023-02-14
[SC-8025] Only close available on manage aave v3

jakubswierczek:
# Changes
- Each strategy can define possible actions on the manage page.
- Earn wsETH/ETH right now support only close operation.


features/aave/common/StrategyConfigTypes.ts
```diff
@@ -1,11 +1,12 @@
 import { IRiskRatio } from '@oasisdex/oasis-actions'
+import { AaveV2ReserveConfigurationData } from 'blockchain/aave'
+import { ManagePositionAvailableActions } from 'features/aave/strategyConfig'
```
sennett, line 3: Circular dependency here.  We could move this type to `IStrategyConfig`.

sennett (approved):

PR #2036, opened 2023-02-15
don't call lib when amount is zero

sennett:
# [don't call lib when amount is zero](https://app.shortcut.com/oazo-apps/story/8069/race-condition-when-entering-small-amount)
  
## Changes 👷‍♀️
- don't call the lib when amount is zero
- prevents race condition where zero value breaks the slider


features/aave/open/state/openAaveStateMachine.ts
```diff
@@ -196,10 +196,21 @@ export function createOpenAaveStateMachine(
...
-                  target: '#openAaveStateMachine.background.debouncing',
-                  actions: ['setAmount', 'calculateAuxiliaryAmount'],
-                },
+                SET_AMOUNT: [
+                  {
+                    target: '#openAaveStateMachine.background.debouncing',
+                    // only call library greater-than-zero amount
+                    cond: (context, event) => {
+                      return !!(event.amount && event.amount.gt(0))
+                    },
```
marcinciarka, line 205: Lets move this to `guards`.

marcinciarka (changes requested):

features/aave/open/state/openAaveStateMachine.ts
sennett: Ah yup - done

marcinciarka (approved):

PR #2042, opened 2023-02-16
[SC-8085] Default slippage in strategy

jakubswierczek:
# Changes
- Introduce a default slippage for the strategy.
- Default slippage for Earn `wstETH/ETH` -> 0.1%.
- All transaction for `wsETH/ETH` has the default slippage set.
- User can change it to theirs (get from user settings).
- New tooltip for Order Information. The tooltip can have action and doesn't disappear after moving the mouse.

### DIFF
--- a/features/aave/common/components/informationContainer/SlippageInformation.tsx
+++ b/features/aave/common/components/informationContainer/SlippageInformation.tsx
@@ -1,15 +1,43 @@
 import BigNumber from 'bignumber.js'
+import { SidebarSectionFooterButton } from 'components/sidebar/SidebarSectionFooterButton'
+import { VaultChangesInformationItem } from 'components/vault/VaultChangesInformation'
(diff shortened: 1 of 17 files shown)


features/aave/common/components/informationContainer/SlippageInformation.tsx
```diff
@@ -18,3 +46,54 @@ export function SlippageInformation({ slippage }: LtvInformationProps) {
...
+            />
+          </OrderInformationTooltipAction>
+        </Flex>
+      }
+    />
+  )
+}
+
+export function SlippageInformation(props: SlippageInformationProps) {
+  if (!props.isStrategyHasSlippage) return <BasicSlippageInformation {...props} />
```
marcinciarka, line 95: Dont like that name, lets change it later (`isStrategyHasSlippage`)

marcinciarka (approved):

PR #2045, opened 2023-02-16
🧿 Separate borrow related stuff into its own context

piotrkonowrocki:
# [🧿 Separate borrow related stuff into its own context](https://app.shortcut.com/oazo-apps/story/8091/earn-context-incorporated-in-borrow-context)

Dedicated proxy for Ajna Borrow. Does not change anything in Borrow reducto or transactions.
  
## Changes 👷‍♀️

- Split existing context into `AjnaBorrowContext` and `AjnaProductContext`,
- adjusted existing files to accommodate this change.
  
## How to test 🧪

See if open and manage flow are still working.

### DIFF
--- a/features/ajna/borrow/contexts/AjnaBorrowContext.tsx
+++ b/features/ajna/borrow/contexts/AjnaBorrowContext.tsx
@@ -1,0 +1,174 @@
+import { AjnaSimulationData } from 'actions/ajna'
+import { useAppContext } from 'components/AppContextProvider'
+import { useGasEstimationContext } from 'components/GasEstimationContextProvider'
+import { ValidationMessagesInput } from 'components/ValidationMessages'
+import { isBorrowFormValid } from 'features/ajna/borrow/contexts/isBorrowFormValid'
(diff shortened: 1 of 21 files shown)


features/ajna/borrow/contexts/AjnaBorrowContext.tsx
```diff
@@ -0,0 +1,177 @@
...
+  children,
+  position,
+}: PropsWithChildren<AjnaBorrowContextProviderProps>) {
+  const { walletAddress } = useAccount()
+  const gasEstimation = useGasEstimationContext()
+  const {
+    environment: { collateralBalance, collateralToken, ethBalance, ethPrice, flow, quoteBalance },
+    steps: { currentStep },
+    tx: { txDetails },
+  } = useAjnaProductContext()
```
piekczyk, line 79: I think that we could skip part where we are defining it as separate context and we could just calculate it on `AjnaProductInput` control level and inject in `AnjaBorrowContext` / `AjnaEarnContext` as props. In that way we could expose these props + context specific values as a single context.

piekczyk (approved):

PR #2070, opened 2023-02-22
Enable Aave SL by default

piekczyk:
# Enable Aave SL by default

## Changes 👷‍♀️

- Aave SL feature flags are set to true
- restored validations
  
## How to test 🧪

- Aave SL should be available without manually setting feature flags to true
- Aave SL validations in overview should be visible


piekczyk:
closed since it's already on prod

PR #2104, opened 2023-02-28
🧿 Ajna content pages improvements

piotrkonowrocki:
# 🧿 Ajna content pages improvements

Couple QOL minor improvements to static pages for Ajna.
  
## Changes 👷‍♀️

- Extracted `AnimatedWrapper` component to be used along all Ajna pages (and also outside of Ajna context, but this PR does not do that),
- extracted `AjnaHeaderProps` component to be used at top of all Ajna pages except HP (or maybe HP includes if it's tweak enough?),
- added support for navigation components so an item in menu can both have it's own dropdown panel and be a link at the same time - also added links for Borrow, Earn and Multiply pages in Ajna navigation,
- fixed couple small inconsistencies with design on rewards page and simplified styles a little bit,
- created empty Ajna Multiply page - this one does not contain product cards as in the light of Oasis create I'm not sure if they are in they final version.
  
## How to test 🧪

No special instructions.

### DIFF
--- a/components/navigation/NavigationMenuPanel.tsx
+++ b/components/navigation/NavigationMenuPanel.tsx
@@ -1,3 +1,4 @@
+import { AppLink } from 'components/Links'
 import React, { ReactNode } from 'react'
 import { Box, Text } from 'theme-ui'
 
@@ -18,6 +19,7 @@ export interface NavigationMenuPanelType {
     label: string
     link: string
   }
+  link?: string
   links: NavigationMenuPanelLink[]
   otherAssets?: NavigationMenuPanelAsset[]
 }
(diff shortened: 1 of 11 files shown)


components/navigation/NavigationMenuPanel.tsx
```diff
@@ -39,26 +42,51 @@ export function NavigationMenuPanel({
...
+          currentPanel={currentPanel}
+          label={label}
+          isPanelOpen={isPanelOpen}
+        />
+      )}
     </Box>
   )
 }
+
+export function NavigationMenuPanelLabel({
```
piekczyk, line 72: could you move declaration of this component above declaration of `NavigationMenuPanel`? I know that hoisting do the job but I found it confusing since files are usually read from top to bottom.

piekczyk (changes requested):

components/navigation/NavigationMenuPanel.tsx
piotrkonowrocki: Fixed.

PR #2145, opened 2023-03-09
🧿 Split get Ajna position observable

piotrkonowrocki:
# 🧿 Split get Ajna position observable

Since library requires quote price, token prices has to loaded first before position is loaded, main observable had to be split into two separated ones. This PR does not pass quote price to library (library is not updated so it doesn't require it) but allows to do it in easy way.
  
## Changes 👷‍♀️

- Remove `dpmPositionData$` observable dependency from `ajnaPosition$` and replaced it with `dpmPositionData` interface,
- created `getStaticDpmPositionData$` fakeish observable,
- updated `AjnaProductController` observable order and dependencies.
  
## How to test 🧪

Try both open and manage flow and see if empty position and existing positions are loading correctly.

### DIFF
--- a/features/ajna/positions/common/observables/getDpmPositionData.ts
+++ b/features/ajna/positions/common/observables/getDpmPositionData.ts
@@ -3,23 +3,22 @@ import { ethers } from 'ethers'
 import { ProxiesRelatedWithPosition } from 'features/aave/helpers/getProxiesRelatedWithPosition'
 import { PositionCreated } from 'features/aave/services/readPositionCreatedEvents'
 import { PositionId } from 'features/aave/types'
-import { AjnaProduct } from 'features/ajna/common/types'
 import { isEqual } from 'lodash'
 import { combineLatest, EMPTY, Observable, of } from 'rxjs'
 import { distinctUntilChanged, map, shareReplay, startWith, switchMap } from 'rxjs/operators'
 
-export interface DpmPositionData extends UserDpmAccount {
+export interface DpmPositionData<P> extends UserDpmAccount {
   collateralToken: string
-  product: AjnaProduct
+  product: P
   protocol: string
   quoteToken: string
 }
 
-export function getDpmPositionData$(
+export function getDpmPositionData$<P>(
   proxiesForPosition$: (positionId: PositionId) => Observable<ProxiesRelatedWithPosition>,
   lastCreatedPositionForProxy$: (proxyAddress: string) => Observable<PositionCreated>,
   positionId: PositionId,
-): Observable<DpmPositionData> {
+): Observable<DpmPositionData<P>> {
   return proxiesForPosition$(positionId).pipe(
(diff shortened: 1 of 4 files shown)


features/ajna/positions/common/observables/getDpmPositionData.ts
```diff
@@ -40,3 +42,25 @@ export function getDpmPositionData$(
...
+}: {
+  collateralToken: string
+  product: AjnaProduct
+  quoteToken: string
+}): Observable<DpmPositionData> {
+  return EMPTY.pipe(
+    startWith({
+      collateralToken,
+      product,
+      protocol: 'Ajna',
```
piekczyk, line 59: you should pass it as parameter if you would like to keep entire method generic and reusable for other protocols as well

features/ajna/positions/common/observables/getDpmPositionData.ts
```diff
@@ -30,7 +32,7 @@ export function getDpmPositionData$(
         ? {
             ...dpmProxy,
             collateralToken: lastCreatedPosition.collateralTokenSymbol,
-            product: lastCreatedPosition.positionType,
+            product: lastCreatedPosition.positionType.toLowerCase() as AjnaProduct,
```
piekczyk, line 35: by this casting you binded entire method to ajna protocol

features/ajna/positions/common/observables/getDpmPositionData.ts
```diff
@@ -1,14 +1,16 @@
...
 import { isEqual } from 'lodash'
-import { combineLatest, Observable, of } from 'rxjs'
-import { distinctUntilChanged, map, shareReplay, switchMap } from 'rxjs/operators'
+import { combineLatest, EMPTY, Observable, of } from 'rxjs'
+import { distinctUntilChanged, map, shareReplay, startWith, switchMap } from 'rxjs/operators'
 
 export interface DpmPositionData extends UserDpmAccount {
   collateralToken: string
-  product: string
+  product: AjnaProduct
```
piekczyk, line 13: I don't think it is correct approach, product may be different for different protocols

piekczyk (changes requested):

features/ajna/positions/common/observables/getDpmPositionData.ts
piotrkonowrocki: Protocol changed to parameter.

features/ajna/positions/common/observables/getDpmPositionData.ts
piotrkonowrocki: Changed to generic.

features/ajna/positions/common/observables/getDpmPositionData.ts
piotrkonowrocki: Changed to generic.

piekczyk (approved):

PR #2171, opened 2023-03-16
[SC-8434] Retry on Getting Parameters on AAVE. Default Page - Earn

jakubswierczek:
# Changes
- Retry getting parameters when gas estimation fails. Mostly we want to ask 1inch about the best route again. 
- Default tab on mainpage: Earn.


BackToTheCode (approved):
Approved for code only. Wasn't able to complete.

PR #2183, opened 2023-03-21
🧿 Predefined min/max values for form inputs

piotrkonowrocki:
# [🧿 Predefined min/max values for form inputs](https://app.shortcut.com/oazo-apps/story/8092/predefined-min-max-values-for-form-inputs)
 
image

## Changes 👷‍♀️

- Added min/max option to all Ajna Borrow form inputs:
  - Deposit:
    - Min: n/a,
    - Max: balance of collateral token in user's wallet.
  - Generate:
    - Min: pool dust limit (considering debt that is already in position),
    - ~~Max: debt available to generate in position before it reaches illegal state (below zero).~~ Temporary turned off.
  - Payback:
    - Min: n/a,
    - Max: balance of quote token in user's wallet or debt left in positions, whichever smaller.
  - Withdraw:
    - Min: n/a,
    - Max: collateral available in position before it reaches illegal state (below zero).'
- Temporary removed min/max from Ajna Earn since it wasn't finished before and interfaces had to be updated, so I will redo min/max for Earn as a separated PR.

Note: some of the min/max values has to be calculated on front-end as they depend on both current position state and simulation state.
  
## How to test 🧪

Try using min/max values in both open and manage flows for Ajna Borrow.

## TODO 📝

As of today, max debt cannot be counted. Initial idea was to subtract from debt available one day of interest, but this is far to low and position still claims to be undercollateralized. I've tested couple scenarios and turns out that lower debt available, the buffer has to be bigger. Calculating that is going to be a part of another story in near future.


features/ajna/positions/common/sidebars/AjnaFormFields.tsx
```diff
@@ -141,16 +185,16 @@ export function AjnaFormFieldPayback({
       action="Payback"
       currencyCode={quoteToken}
       tokenUsdPrice={quotePrice}
-      amount={state.paybackAmount}
-      auxiliaryAmount={state.paybackAmountUSD}
+      amount={state.paybackAmount?.decimalPlaces(6)}
+      auxiliaryAmount={state.paybackAmountUSD?.decimalPlaces(6)}
```
piekczyk, line 189: we should use here quotePrecision

features/ajna/positions/borrow/helpers/getAjnaBorrowCollateralMax.ts
```diff
@@ -0,0 +1,25 @@
...
+export function getAjnaBorrowCollateralMax({
+  digits,
+  position: {
+    collateralAmount,
+    debtAmount,
+    pool: { lowestUtilizedPrice },
+  },
+  simulation,
+}: AjnaBorrowCollateralMaxParams) {
+  const resolveedDebtAmount = simulation?.debtAmount || debtAmount
```
piekczyk, line 19: typo -`resolvedDebtAmount`

piekczyk (changes requested):

features/ajna/positions/borrow/helpers/getAjnaBorrowCollateralMax.ts
piotrkonowrocki: Fixed.

PR #2186, opened 2023-03-22
:fire: [SC-8513] Remove /connect page

jakubswierczek:
# Changes
- removed `/connect` page
- removed all old connectors
- removed a package from `Dockerfile` which was used to connect HW
- removed `UseBlocknativeOnboard` feature flag.


features/web3Context/web3_context.ts
```diff
@@ -198,7 +131,7 @@ export function createWeb3Context$(
           connectionKind,
           web3: library as any,
           chainId: chainId!,
-          account: ['ledger', 'trezor'].indexOf(connectionKind) >= 0 ? hwAccount! : account,
+          account: account,
```
piekczyk, line 134: account,

Dockerfile
```diff
@@ -9,7 +9,6 @@ COPY ./blockchain/abi/*.json /usr/src/app/blockchain/abi/
 
 WORKDIR /usr/src/app
 
-RUN apt update && apt-get install -y libudev-dev && apt-get install libusb-1.0-0
```
piekczyk, line 12: I'm confused by this one, could you elaborate?

piekczyk (changes requested):

Dockerfile
jakubswierczek: Yes, as we move away from our connection methods, we don't need to connect hardware wallets directly. We will be using LedgerLive instead of speaking to the USB device.

piekczyk (approved):

PR #2216, opened 2023-04-03
Ajna product card links

piekczyk:
# Ajna product card links

## Changes 👷‍♀️

- added links to product cards
- prepared initial version of ajna product cards observable which in next step will be extended with data needed for displaying stuff on UI per specific product card type
- extracted few common components

  
## How to test 🧪

- product cards available on Ajna lending page should redirect to proper pool selector


features/ajna/common/components/AjnaProductCardsWrapper.tsx
```diff
@@ -0,0 +1,12 @@
+import { ProductCardsWrapper } from 'components/productCards/ProductCardsWrapper'
+import React, { ReactNodeArray } from 'react'
+
+interface AjnaProductCardsWrapperProps {
+  children: ReactNodeArray
+}
+
+export const AjnaProductCardsWrapper = ({ children }: AjnaProductCardsWrapperProps) => (
+  <ProductCardsWrapper sx={{ mt: ['9px', '48px'], gap: ['88px', 3, 3] }}>
```
piotrkonowrocki, line 9: `9px` sounds kinda magic numbery, that should probably be `8px`, so just `2` from array.

features/ajna/common/content.ts
```diff
@@ -1,5 +1,24 @@
 import BigNumber from 'bignumber.js'
 import { AjnaPoolData } from 'features/ajna/common/types'
+import { INTERNAL_LINKS } from 'helpers/applicationLinks'
+
+type AjnaProductCardBorrowBase = {
```
piotrkonowrocki, line 5: 1. Is this type anyhow Ajna specific? Could this component and it's type be reused in other places?
2. Even if not, file named `content.ts` doesn't sound like a good place to store type.

features/ajna/common/content.ts
```diff
@@ -1,5 +1,24 @@
...
+  }
+  button: {
+    link: string
+    hash: string
+    labelKey: string
+  }
+  labels: { titleKey: string; value: string }[]
+}
+
+type AjnaProductCardEarnBase = AjnaProductCardBorrowBase
```
piotrkonowrocki, line 21: Yet another reason to make it more generic since both borrow and earn are using same type. Just `AlternateProductCardBase`.

features/ajna/common/content.ts
```diff
@@ -90,3 +145,32 @@ export const ajnaPoolDummyData: AjnaPoolData = {
     tvl: new BigNumber(Math.random() * 10000000),
   },
 }
+
+export const benefitCardsAnja = [
```
piotrkonowrocki, line 149: Suggestion, let me know what you think. This is something that I actually wanted to talk about on front-end conventions panel, but time did not allowed to do that.

What about changing approach to design pattern with components in views? Instead of having this config in separated place and passing it directly to abstract component on landing page view, create controller. Controller doesn't have any props, you could include it directly in view. Then, inside of controller you could store those config directly since it's not abstract, it's Ajna specific, that one single place specific, and then pass it to abstract component.

~~Same could be applied to product cards config.~~ Now I noticed that product cards already have their controller working similar to what I just described, to that would be a better place to store all those data and its types and get rid of this file.

Happy to talk more about it more.

patches/@oasisdex+oasis-actions+0.2.16.patch
piotrkonowrocki, line 1: Is that change related to this PR and intentional?

piotrkonowrocki (changes requested):

patches/@oasisdex+oasis-actions+0.2.16.patch
piekczyk: yes, it is some kind of left over

piotrkonowrocki (approved):

PR #2218, opened 2023-04-03
Make Lending Protocol more usable

jakubswierczek:
# Changes
- Right now `Lending Protocol` enum can be used everywhere in our app.
- Introduced `AaveServices`, an interface for Aave v2 and Aave v3.
- Tests are still not working, I installed a new library for it, and added a configuration, but it needs more work. 
- Refactor Aave stuff.


package.json
```diff
@@ -121,7 +121,7 @@
     "text-mask-addons": "^3.7.2",
     "theme-ui": "^0.3.1",
     "ts-essentials": "^7.0.1",
-    "ts-node": "^7.0.1",
+    "ts-node": "10.9.1",
```
marcinciarka, line 124: We could use a jest script in here.

marcinciarka (approved):

PR #2220, opened 2023-04-06
Network switcher fixes

marcinciarka:
# [Network switcher fixes](https://app.shortcut.com/oazo-apps/story/8644/network-switcher-fixes)

## Changes 👷‍♀️
- styling fixes (tick position, hover animation)
- synced settings when the network is forced (eg. on reconnect), this fixed the blicking connect button
- added a redirection of user closes the connection modal on a page where connection is needed (or disconnects on such page)

## How to test
- the network switcher should look better, should have hover effects
- the errors where you access a page which requires a wallet connection (eg. `vaults/open-multiply/WSTETH-A`) with no wallet connected should be gone
- basically everything should be fine just fine :v:


components/connectWallet/Connection.tsx
```diff
@@ -1,15 +1,29 @@
...
-import { useEffect } from 'react'
+import { useRouter } from 'next/router'
+import { useCallback, useEffect } from 'react'
 
 export function Connection({ children, walletConnect }: WithChildren & { walletConnect: boolean }) {
   const { executeConnection, connected, connecting } = useWeb3OnBoardConnection({
     walletConnect,
   })
+  const { replace, reload } = useRouter()
+  const handleConnected = useCallback(
```
jakubswierczek, line 12: We can await a callback, so let's do so for `replace`.

components/navigation/NavigationPickNetwork.tsx
```diff
@@ -105,8 +110,11 @@ export function NavigationPickNetwork() {
                       sx={{
                         width: '100%',
                         textAlign: 'right',
+                        position: 'relative',
                         opacity: isCurrentNetwork ? 1 : 0,
-                        transition: '0.2s opacity',
+                        left: isCurrentNetwork ? 0 : '5px',
```
jakubswierczek, line 115: Magic 5px?

jakubswierczek (changes requested):

components/connectWallet/Connection.tsx
marcinciarka: Fixed.

components/navigation/NavigationPickNetwork.tsx
marcinciarka: Fixed.

jakubswierczek (approved):

PR #2223, opened 2023-04-11
🧿 Extend Ajna Context for Multiply Product

piotrkonowrocki:
# [Extend Ajna Context for Multiply Product](https://app.shortcut.com/oazo-apps/story/8660/extend-ajna-context-for-multiply-product)

Extended `AjnaProductContext` and `AjnaProductController` to fit Multiply product. So far this is just the context, no UI changes were made.
  
## Changes 👷‍♀️

- Created fake `AjnaMuliplyPosition` interface to be replaced in the future with one from library,
- added support for Multiply product in Ajna context,
- created `AjnaMultiplyFormReducto` for opening part of the flow.
  
## How to test 🧪

- See if `/ajna/multiply/WBTC-USDC` doesn't break the application,
- see if Borrow/Earn flows are still working.

### DIFF
--- a/features/ajna/positions/common/contexts/AjnaProductContext.tsx
+++ b/features/ajna/positions/common/contexts/AjnaProductContext.tsx
@@ -16,6 +16,11 @@ import {
   AjnaEarnFormState,
   useAjnaEarnFormReducto,
 } from 'features/ajna/positions/earn/state/ajnaEarnFormReducto'
+import {
+  AjnaMultiplyFormState,
+  useAjnaMultiplyFormReducto,
+} from 'features/ajna/positions/multiply/state/ajnaMultiplyFormReducto'
+import { AjnaMultiplyPosition } from 'features/ajna/positions/multiply/temp'
 import { useObservable } from 'helpers/observableHook'
 import { useAccount } from 'helpers/useAccount'
 import React, {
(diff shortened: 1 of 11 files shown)


features/ajna/positions/common/contexts/AjnaProductContext.tsx
```diff
@@ -87,19 +91,33 @@ type AjnaProductContextWithEarn = AjnaProductContext<
...
+  ? AjnaProductContextWithEarn
+  : T extends 'multiply'
+  ? AjnaProductContextWithMultiply
   : never
 
 export function useAjnaProductContext<T extends AjnaProduct>(product: T): PickProductType<T> {
   const { environment } = useAjnaGeneralContext()
-  const context = product === 'borrow' ? useContext(ajnaBorrowContext) : useContext(ajnaEarnContext)
+  const context =
+    product === 'borrow'
```
piekczyk, line 116: maybe instead of this we could use map? 

```
const context = {
   borrow: useContext(ajnaBorrowContext),
   earn: useContext(ajnaEarnContext),
   multiply: useContext(ajnaMultiplyContext),
}
```

features/ajna/positions/common/controls/AjnaProductController.tsx
```diff
@@ -215,6 +217,18 @@ export function AjnaProductController({
                             <AjnaEarnPositionController />
                           </AjnaProductContextProvider>
                         )}
+                        {dpmPosition.product === 'multiply' && (
```
piekczyk, line 220: It won't work for cases where product type was overwritten by user while switching position type on UI. Here we need to use data from borrow db.

piekczyk (changes requested):

features/ajna/positions/common/contexts/AjnaProductContext.tsx
piotrkonowrocki: Then in the end of this method we would have to check which of those context from map has to be returned. This condition is inevitable, but if we use it in the beginning, at least we are creating only the context we actually need, not all three of them.

features/ajna/positions/common/controls/AjnaProductController.tsx
piotrkonowrocki: Fixed, product type is checked on observable level.

piekczyk (approved):

PR #2227, opened 2023-04-12
Ajna mining rewards

piekczyk:
# [Ajna mining rewards](https://app.shortcut.com/oazo-apps/story/8562/fe-ajna-lm-reward-claiming)

## Changes 👷‍♀️

- updated designs
- added transaction handling (not tested yet)
- added data source
- fixed issue with vault api call using protocol with capital first letter
  
## How to test 🧪

- test whether designs are in line with figma (there will be minor differences due to agreement to not use magic numbers)
- check content of rewards page when connected and not connected (when connected banner in card should be displayed, if not connected banner should be hidden and connect button in page header should be visible)
- ajna token rewards card for now doesn't have banner because is not supported yet
- transaction handling is not testable due to lack of contract


features/ajna/rewards/useAjnaUserNfts.ts
```diff
@@ -0,0 +1,90 @@
...
+        })
+        .catch((error) => {
+          console.error(error)
+          setIsLoading(false)
+        })
+    }
+  }, [walletAddress, txDetails?.txStatus])
+
+  if (!nfts?.length) {
+    return {
```
piotrkonowrocki, line 52: Consider if it's doable to merge those two returns into a one response. They are both returning same interface, but imo would be easier to read instead of splitting depending on single condition.

If it's not possible, then just resolve this comment.

piotrkonowrocki (changes requested):
Comments about UI:

1. Missing "Exclusive to Oasis.app" pill.
image
2. I really don't like skeleton loading in place of entire box with rewards details, but that's just my personal opinion. I would prefer to display separated smaller animations in place of every value that isn't available from beginning. But on the other hand, that might look bad because of gradient background underneath it. I suggest getting some opinions about that from Hazel.
3. There are some minor deviations from design in terms of spacing in reward cards, but I think they are reasonable and there is nothing that might require reworking.

components/WithArrow.tsx
```diff
@@ -23,6 +29,7 @@ export function WithArrow({
         },
         ...sx,
       }}
+      {...(as ? { as } : {})}
```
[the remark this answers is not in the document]
piekczyk: fixed

features/ajna/rewards/useAjnaUserNfts.ts
piekczyk: I've done that in this way because I was able to define `tokenRewards` and use it in two places without making any additional conditions.

piotrkonowrocki (approved):

PR #2269, opened 2023-05-04
Refine and rename Discover table

piotrkonowrocki:
# Refine and rename Discover table

Partially covers and is required for: https://app.shortcut.com/oazo-apps/story/8904/oasis-create-add-sorting-by-column-to-discover-table

This PR is a lot smaller than it initially looks. Most of the changes are just file renames (for some reason github threats some of them as new files 🤔) or splitting big files in smaller chunks.
  
## Changes 👷‍♀️

- Renamed `DiscoverTable` component to `AssetsTable` along with all partial components that it uses (`DiscoverCards` -> `AssetsCards`) - for unknown reason in some cases github doesn't recognize renaming/moving file and marks it as a new file,
- moved `AssetsTable` component and its dependencies from `discover` directory to `components`,
- removed all Oasis Discover specific properties from `AssetsTable` to make the component completely abstract,
- removed old way of rendering table data - from now table accepts just a value as a string, number or node and display it as is without formatting that was Discover specific - all data has to be preformatted before passing it to the component,
- similar thing applies to the table on owner's page as it was utilizing non-generic Discover formatting,
- split old `DiscoverTableDataCellContent.tsx` file into smaller chunks, separated for each component into `components/assetsTable/cellComponents` directory,
- split old helper file into smaller chunks, separated for each functions - some of them became a part of abstract `AssetsTable` component, some of them were left as a part of Discover feature.
  
## How to test 🧪

See if following pages works:
- Discover,
- owner's page, preferable on multiple addresses with different types of positions on different protocols,
- Ajna pool selector.

Those are all of the pages that uses `AssetsTable` (prev. `DiscoverTable`).


analytics/analytics.ts
```diff
@@ -2,7 +2,7 @@ import { ProductType } from 'analytics/common'
 import BigNumber from 'bignumber.js'
 import { Context } from 'blockchain/network'
 import browserDetect from 'browser-detect'
-import { getDiscoverMixpanelPage } from 'features/discover/helpers'
+import { getDiscoverMixpanelPage } from 'features/discover/helpers/getDiscoverMixpanelPage'
```
jakubswierczek, line 5: Can we create an index file?

components/assetsTable/types.tsx
jakubswierczek, line 1: Please create index files in `assetsTable` and change names to kebab case

public/locales/en/common.json
jakubswierczek, line 1: Add the key to other languages

jakubswierczek (commented):
Overall it’s good work, but it needs adjustment to our golden rules. It’s a great opportunity to add index files and change names to kebab case

jakubswierczek (approved):

PR #2273, opened 2023-05-09
🧬 Natural language selector

piotrkonowrocki:
# [Natural language selector](https://app.shortcut.com/oazo-apps/story/8916/update-natural-language-selector-dropdown)

Example page with natural language selector for Oasis Create.

image
  
## Changes 👷‍♀️

- Created temp page `oasis-create` for testing and experiments - the page is hidden behind feature flag,
- created `NaturalLanguageSelectorController` - for now it's filled with fake data that should to be replaced with data taken from API when it's ready,
- updated visuals of `HeaderSelector`,
- does not include balances in second select as design is not ready for them.
  
## How to test 🧪

- Turn on `OasisCreate` feature flag,
- go to `oasis-create/borrow` and change something in the filters.

Please note that updating first selector changes URL. As it works correctly in project running as production build, during testing in dev build there are long delays between page reloads as NextJS rebuilds page every time. To test it properly please build project first then run it. This is similar thing as switching tabs in Discover.


marcinciarka (approved):
👌

PR #2291, opened 2023-05-18
Ad tenderly as rpc

adamskrodzki:
# [Adjust /rpc endpoint to be switchable to tenderly fork]
https://app.shortcut.com/oazo-apps/story/8981/adjust-rpc-endpoint-to-be-switchable-to-tenderly-fork

**THIS IS 1/3 PR's see full epic for context. ENV variable approach is temporary and will be replaced with database.**

  
## Changes 👷‍♀️
- new environment variable TENDERLY_FORK_ID  in format like b12b2b04-f54b-11ed-a05b-0242ac120003
- feature flag UseTenderlyRPC 
- if feature flag is set to true all RPC traffic is redirected to tenderly

  
## How to test 🧪

IN order to test it properly you need also set either tenderly or oasis rpc url as your RPC in a wallet

At that point any functionality that is not 1 inch dependent should work (draw debt for example) across whole aplication


paszkowskiDamian (changes requested):
I would go with earn solution where you can change rpc on the UI level. @jakubswierczek

jakubswierczek (changes requested):
We shouldn't use Tenderly explicitly in our app. We are working on a solution to provide a custom RPC node in the network switcher.
Tenderly isn't a network. It's a provider. 
Close the pull request and never bring it up again.

jakubswierczek (changes requested):
We shouldn't use Tenderly explicitly in our app. We are working on a solution to provide a custom RPC node in the network switcher.
Tenderly isn't a network. It's a provider. 
Close the pull request and never bring it up again.

marcinciarka:
+1

adamskrodzki:
@jakubswierczek @marcinciarka We need a way to test ajna end-to-end with 1inch integration. Read epic related to the story and feel free to propose better solution, but ensure you understodd epic first. 

Problem is:

1) We need nodes that sync quickly (so hardhat is off the table)
2) We need node forked from relatively fresh block, so pasting url like https://rpc.tenderly.co/fork/[YOUR HARDCODED GUID HERE] is off the table.

we need some kind of proxy to change forks in a background.

pages/api/rpc.tsx
```diff
@@ -61,7 +61,11 @@ const blockRecheckDelay = 3000
 
 const cache: { [key: string]: Cache } = {}
 
-function getRpcNode(network: NetworkNames) {
+function getRpcNode(network: NetworkNames, tenderlySecret: string) {
+  //TODO: take Fork_ID from database and refresh it every 5 minutes
+  if (tenderlySecret === process.env.TENDERLY_SECRET && network === NetworkNames.ethereumTenderly) {
```
adamskrodzki, line 66: authorization,
to prevent random clients to use our tenderly. requires to set correct secret value in local storage

.env
```diff
@@ -14,6 +14,12 @@ ADROLL_PIX_ID="4PSXWPAQTJEIXJSPHJCRW4"
 # This is a test value for local development
 INFURA_PROJECT_ID_BACKEND="de82b2d602264e4fbc0929dec0c45baa"
 ETHERSCAN_API_KEY="34JVYM6RPM3J1SK8QXQFRNSHD9XG4UHXVU"
+
+# This is test value that do not work, if you want to test it locally, please use your own. To do that create Tenderly fork and use its id.
+TENDERLY_FORK_ID=b12b2b04-f54b-11ed-a05b-0242ac120003
```
jakubswierczek, line 19: Add empty string here.

components/AppContext.ts
```diff
@@ -127,6 +127,7 @@ import {
   Vault,
 } from 'blockchain/vaults'
 import { pluginDevModeHelpers } from 'components/devModeHelpers'
+import { getTenderlyRpcUrl } from 'config/rpcConfig'
```
jakubswierczek, line 130: I prefer to avoid naming specific products. Can we call it `getCustomForkRpcUrl`?

.env
```diff
@@ -14,6 +14,12 @@ ADROLL_PIX_ID="4PSXWPAQTJEIXJSPHJCRW4"
...
+# must equal to local value in storage (tenderlySecret) if you want to use tenderly fork as rpc provider
+TENDERLY_SECRET="test"
```
jakubswierczek, line 21: Avoid using the name "Tenderly."

components/AppContextProvider.tsx
```diff
@@ -26,10 +27,11 @@ export function useAppContext(): AppContext {
 
 export function AppContextProvider({ children }: WithChildren) {
   const [context, setContext] = useState<AppContext | undefined>(undefined)
+  const tenderlySecret = useLocalSetting('tenderlySecret')
 
   useEffect(() => {
-    setContext(setupAppContext())
-  }, [])
+    setContext(setupAppContext(tenderlySecret))
+  }, [tenderlySecret])
```
jakubswierczek, line 34: We should not refresh the app context more than once, as our entire app relies on it. If we change it, the app will re-render. We should address this issue, but for now, we should simply refresh the page.

components/AppContext.ts
```diff
@@ -541,9 +542,12 @@ function initializeUIChanges() {
   return uiChangesSubject
 }
 
-export function setupAppContext() {
+export function setupAppContext(tenderlySecret: string) {
```
jakubswierczek, line 545: I have a negative opinion about including parameters in the app context initialiser.

components/AppContext.ts
```diff
@@ -541,9 +542,12 @@ function initializeUIChanges() {
...
   const once$ = of(undefined).pipe(shareReplay(1))
   const chainIdToRpcUrl = mapValues(networksById, (network) => network.rpcUrl)
+  if (tenderlySecret) {
+    chainIdToRpcUrl[NetworkIds.MAINNET] = getTenderlyRpcUrl(tenderlySecret)
```
jakubswierczek, line 549: Try not to alter the state.

jakubswierczek (changes requested):

.env
adamskrodzki: removed from TENDERLY_SECRET  but leaving in TENDERLY_FORK_ID

components/AppContextProvider.tsx
adamskrodzki: ok, I will change it back to [], my reasoning was that You want to refresh whole application if that value changes. But on other hand tenderlySecret is not reactive anyway

components/AppContext.ts
adamskrodzki: I do not have strong opinion on that matter I can add
`const tenderlySecret = getStorageValue<string>('tenderlySecret', '')` inside setupAppContext() 

but I feel that it's better to have in AppContextProvider since it is smaller AppContext so it is more visible, that we depend on some storage value

components/AppContext.ts
adamskrodzki: what do you mean? that is a place where `chainIdToRpcUrl ` is being created
do you want me to extract those 2 lines to a function ? i think that's overshooting

helpers/useFeatureToggle.ts
```diff
@@ -121,6 +123,14 @@ export function getFeatureToggle(feature: Feature): boolean {
   return false
 }
 
+export function useLocalSetting(setting: Setting): string {
```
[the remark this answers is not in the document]
adamskrodzki: done

.env
adamskrodzki: I do not understand this comment

PR #2294, opened 2023-05-18
🧿 Allow to set price above MOMP in Ajna Earn position

piotrkonowrocki:
# [Allow to set price above MOMP in Ajna Earn position](https://app.shortcut.com/oazo-apps/story/9141/earn-can-t-deposit-above-momp)

Bugfix.
  
## Changes 👷‍♀️

- Validation of price above MOMP is now warning, not error,
- fixed visually disabling slider component,
- bonus: slider is disabled until user input something in deposit field.
  
## How to test 🧪

Self explanatory.

### DIFF
--- a/features/ajna/positions/earn/sidebars/AjnaEarnFormContentOpen.tsx
+++ b/features/ajna/positions/earn/sidebars/AjnaEarnFormContentOpen.tsx
@@ -11,7 +11,10 @@ export function AjnaEarnFormContentOpen() {
     environment: { quotePrice, quoteToken },
   } = useAjnaGeneralContext()
   const {
-    form: { dispatch },
+    form: {
+      dispatch,
+      state: { depositAmount },
+    },
     validation: { isFormValid },
   } = useAjnaProductContext('earn')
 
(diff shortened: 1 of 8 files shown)


features/ajna/positions/earn/sidebars/AjnaEarnFormContentOpen.tsx
```diff
@@ -23,7 +26,7 @@ export function AjnaEarnFormContentOpen() {
         tokenPrice={quotePrice}
         resetOnClear
       />
-      <AjnaEarnSlider />
+      <AjnaEarnSlider isDisabled={!depositAmount || depositAmount?.lte(0)} />
```
piekczyk, line 29: not sure if it makes sense to pass it as prop since we extensively use Ajna Product and General context inside of `AjnaEarnSlider` component

piekczyk (changes requested):
general comment - since we change price above momp validation from error to warning, I guess we should also update notification

features/ajna/positions/earn/sidebars/AjnaEarnFormContentOpen.tsx
piotrkonowrocki: But this is the only case, where this slider has to be disabled. In any other cases on manage form, deposit + adjust or withdraw empty deposit doesn't block it, the calculation works.

piekczyk (approved):

PR #2298, opened 2023-05-19
🧿 Modals for Ajna Borrow

piotrkonowrocki:
# [Modals for Ajna Borrow](https://app.shortcut.com/oazo-apps/story/8965/borrow-add-modals)

Only modals, no tooltips as there must have been a misunderstanding with tooltips - their content and titles doesn't match what we currently have in footer so they are probably are supposed to be in some other place, than we anticipated while estimating the story.

image
  
## Changes 👷‍♀️

- Added modals to overview section cards for Ajna Borrow,
- created helper component `DetailsSectionContentSimpleModal` for all default looking modals* - that allowed to get rid of a lot of duplication previously needed to handle all modals,
- created Ajna specific helper `AjnaDetailsSectionContentSimpleModal` for `DetailsSectionContentSimpleModal` that wraps modals in Ajna theme and applies appropriate colors.

* `ContentCardPositionLendingPrice` is an example of card then modal is more complicated than default. In that case, separated component for modal content is still needed, and `DetailsSectionContentSimpleModal` serves just as a wrapper.
  
## How to test 🧪

- See modals on any Ajna Borrow position,
- see if preexisting modals on Ajna Earn positions didn't break.

### DIFF
--- a/components/DetailsSectionContentSimpleModal.tsx
+++ b/components/DetailsSectionContentSimpleModal.tsx
@@ -1,0 +1,34 @@
+import React, { PropsWithChildren } from 'react'
+import { Card, Grid, Heading, Text } from 'theme-ui'
+
+export interface DetailsSectionContentSimpleModalProps {
+  description?: string
+  title: string
+  value?: string
+}
+
+export function DetailsSectionContentSimpleModal({
+  children,
+  description,
+  title,
+  value,
+}: PropsWithChildren<DetailsSectionContentSimpleModalProps>) {
+  return (
+    <Grid gap={2}>
+      <Heading variant="header5" sx={{ fontWeight: 'bold' }}>
+        {title}
+      </Heading>
+      {description && (
+        <Text variant="paragraph3" as="p" sx={{ color: 'neutral80' }}>
+          {description}
+        </Text>
+      )}
+      <>{children}</>
+      {value && (
+        <Card variant="vaultDetailsCardModal" sx={{ mt: 2 }}>
+          {value}
+        </Card>
+      )}
+    </Grid>
+  )
+}
--- a/components/DetailsSectionContentSimpleModal.tsx
+++ b/components/DetailsSectionContentSimpleModal.tsx
@@ -1,34 +1,32 @@
-import React, { PropsWithChildren } from 'react'
+import React, { FC } from 'react'
 import { Card, Grid, Heading, Text } from 'theme-ui'
 
 export interface DetailsSectionContentSimpleModalProps {
-  description?: string
   title: string
+  description?: string
   value?: string
 }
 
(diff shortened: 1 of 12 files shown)


components/DetailsSectionContentSimpleModal.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { PropsWithChildren } from 'react'
+import { Card, Grid, Heading, Text } from 'theme-ui'
+
+export interface DetailsSectionContentSimpleModalProps {
+  description?: string
+  title: string
```
piekczyk, line 6: IMO it's much nicer if required props are first in the interface, not required after.

piekczyk: Take a look how hints looks when optional parameter is declared first and required one later vs opposite.

image

vs

image

features/ajna/common/components/AjnaDetailsSectionContentSimpleModal.tsx
```diff
@@ -0,0 +1,15 @@
+import {
+  DetailsSectionContentSimpleModal,
+  DetailsSectionContentSimpleModalProps,
+} from 'components/DetailsSectionContentSimpleModal'
+import React, { PropsWithChildren } from 'react'
+import { ajnaExtensionTheme } from 'theme'
+import { ThemeProvider } from 'theme-ui'
+
+export const AjnaDetailsSectionContentSimpleModal = (
```
piekczyk, line 9: What about utilizing `FC`? Btw. when using FC you don't need to use `PropsWithChildren`.

components/DetailsSectionContentSimpleModal.tsx
```diff
@@ -0,0 +1,34 @@
...
+  value?: string
+}
+
+export function DetailsSectionContentSimpleModal({
```
piekczyk, line 10: maybe it's worth to convert this component to arrow function with `FC`

piekczyk (changes requested):

components/DetailsSectionContentSimpleModal.tsx
piotrkonowrocki: Fixed.

components/DetailsSectionContentSimpleModal.tsx
piotrkonowrocki: And fixed.

piekczyk (approved):

features/ajna/common/components/AjnaDetailsSectionContentSimpleModal.tsx
piotrkonowrocki: Fixed.

PR #2322, opened 2023-05-26
Ajna nft opt in/out

piekczyk:
# [Ajna nft opt in/out](https://app.shortcut.com/oazo-apps/story/9236/earn-opt-out-of-nft)

## Changes 👷‍♀️

- added nft step to ajna earn open flow where user can opt in/out from staking nft
  
## How to test 🧪

- in ajna open flow verify whether new step works as expected as well as all transitions (go back to editing, when dpm & allowance is ready, when only dpm  is ready etc.)
- double check if manage earn flow remain untouched

### DIFF
--- a/features/ajna/positions/earn/state/ajnaEarnFormReducto.ts
+++ b/features/ajna/positions/earn/state/ajnaEarnFormReducto.ts
@@ -17,6 +17,7 @@ export interface AjnaEarnFormState {
   price?: BigNumber
   withdrawAmount?: BigNumber
   withdrawAmountUSD?: BigNumber
+  nftOpt?: boolean
   uiDropdown: AjnaEarnPanel
   uiPill: Exclude<AjnaEarnAction, 'open-earn' | 'claim-earn'>
 }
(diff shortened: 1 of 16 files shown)


features/ajna/positions/earn/state/ajnaEarnFormReducto.ts
```diff
@@ -17,6 +17,7 @@ export interface AjnaEarnFormState {
   price?: BigNumber
   withdrawAmount?: BigNumber
   withdrawAmountUSD?: BigNumber
+  nftOpt?: boolean
```
piotrkonowrocki, line 20: Would be better if we could stay consistent and use the same name, as is used in lib: `isStakingNft`.

piotrkonowrocki (changes requested):

features/ajna/positions/earn/state/ajnaEarnFormReducto.ts
piekczyk: fixed

piotrkonowrocki (approved):

PR #2334, opened 2023-05-29
[Draft]add tenderly fork refresh

adamskrodzki:
Implementation of 
https://www.notion.so/Custom-forks-in-FE-app-7b0ffda5286f4abc90390b89bde8b333


halaprix (commented):
needs to get the fix for the `getLogs` issue before it's usable.

halaprix (changes requested):
needs to get the fix for the `getLogs` issue before it's usable. 
also would be good to increase the fork time and add an endpoint to manually refresh the fork in case there are problems with swaps.

PR #2403, opened 2023-06-20
Adding GHA workflow for new prod infra

cristidas:
- Add new GH Actions workflow for the new prod infra

### DIFF
--- a/.github/workflows/new-aws-prod.yaml
+++ b/.github/workflows/new-aws-prod.yaml
@@ -1,0 +1,91 @@
+on:
+  push:
+    branches:
+      - TECH-1747-gha-prod-updates
+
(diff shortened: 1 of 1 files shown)


.github/workflows/new-aws-prod.yaml
```diff
@@ -0,0 +1,91 @@
...
+
+name: Deploy to AWS Production
+
+jobs:
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-latest
+    env:
+      AWS_REGION: us-east-1
+      ECR_REPO_NAME: ${{ secrets.BASE_NAME }}-prod
```
eskp, line 14: Can we make this a var please

PR #2465, opened 2023-06-29
🧿 Ajna safety switch

piotrkonowrocki:
# [Ajna safety switch](https://app.shortcut.com/oazo-apps/story/9999/ajna-safety-switch)

A safety switch that can be turn on to prevent users from dumping tokens in Ajna.
  
## Changes 👷‍♀️

Whenever `AjnaSafetySwitch` is turned on:
- All Ajna open flow forms are locked, primary button is hidden, there is special notification displayed at the top of overview section,
- empty borrow positions are preventing user from depositing more collateral or generating or paying back debt (which isn't a case since position is empty, but still message will appear),
- empty earn positions are preventing user from depositing more quote,
- all Ajna products and promo cards are hidden from Product Finder.
  
## How to test 🧪

Turn on `AjnaSafetySwitch` and go through the list.

### DIFF
--- a/features/ajna/positions/common/validation.tsx
+++ b/features/ajna/positions/common/validation.tsx
@@ -268,6 +268,17 @@ export function getAjnaValidation({
         message: { component: <AjnaSafetyOnMessage /> },
       })
     }
+    if (
+      product === 'earn' &&
+      'quoteTokenAmount' in position &&
+      position.quoteTokenAmount?.isZero() &&
+      'depositAmount' in state &&
+      state.depositAmount?.gt(zero)
+    ) {
+      localErrors.push({
+        message: { component: <AjnaSafetyOnMessage /> },
+      })
+    }
   }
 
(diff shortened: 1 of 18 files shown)


features/ajna/positions/common/validation.tsx
```diff
@@ -244,6 +258,32 @@ export function getAjnaValidation({
     localErrors.push({ message: { translationKey: 'payback-amount-exceeds-debt-token-balance' } })
   }
 
+  if (ajnaSafetySwitchOn && flow === 'manage') {
+    if (
+      product === 'borrow' &&
```
piekczyk, line 263: I think we could extend this condition with multiply case, so beside parameters below we would need to check ltv as well

piekczyk (changes requested):

features/ajna/positions/common/validation.tsx
piotrkonowrocki: Fixed.

PR #2483, opened 2023-07-04
🧿 Update Ajna position URLs

piotrkonowrocki:
# Update Ajna position URLs

Partially covers: https://app.shortcut.com/oazo-apps/story/10023/prevent-actions-on-other-networks
  
## Changes 👷‍♀️

- Moved Ajna open flow to `/ethereum/ajna/{{id}}`,
- moved Ajna manage flow to `/ethereum/ajna/{{product}}/{{pool}}`,
- created redirect for old Ajna position pages.
  
## How to test 🧪

Try open flow and manage flow.


pages/ajna/position/[id].tsx
```diff
@@ -1,27 +1,28 @@
...
+      },
+    }
+  }
 
-export async function getServerSideProps({ locale, query }: GetServerSidePropsContext) {
   return {
-    props: {
-      ...(await serverSideTranslations(locale || 'en', ['common'])),
-      id: query.id,
+    redirect: {
```
Comment (piekczyk) [inline pages/ajna/position/[id].tsx:23]: general comment - do we need this redirect at all since ajna was not released yet?

pages/[networkOrProduct]/ajna/[...pool].tsx
```diff
@@ -0,0 +1,62 @@
...
+  )
+}
+
+AjnaOpenPositionPage.layout = AjnaLayout
+AjnaOpenPositionPage.seoTags = ajnaPageSeoTags
+
+export default AjnaOpenPositionPage
+
+export async function getServerSideProps({ locale, query }: GetServerSidePropsContext) {
+  const network = query.networkOrProduct as string
```
Comment (piekczyk) [inline pages/[networkOrProduct]/ajna/[...pool].tsx:35]: I went through all files and it seems that we always use `networkOrProduct` in terms of network. Maybe `[network]/ajna ...` is a better idea

piekczyk (changes requested):

pages/ajna/position/[id].tsx
Comment (piotrkonowrocki) [reply pages/ajna/position/[id].tsx:23]: I did that for our own convenience, so our old addresses will still work. We can remove it before the release.

pages/[networkOrProduct]/ajna/[...pool].tsx
Comment (piotrkonowrocki) [reply pages/[networkOrProduct]/ajna/[...pool].tsx:35]: This is much more complicated. This directory is named `networkOrProduct`, because it is used for more than one purpose, inside there are files related to products (`/maker`, `/aave` etc.) and to product finder (`/borrow`, `/earn` etc.).

piekczyk (approved):

PR #2508, opened 2023-07-11
Hide the stop loss tabs for every Aave product except supported ones (V3 mainnet)

marcinciarka:
# [Hide the tabs for Aave v2 & Optimism (and arbitrum if that's also set)](https://app.shortcut.com/oazo-apps/story/10241/hide-the-tabs-for-aave-v2-optimism-and-arbitrum-if-that-s-also-set)

  
## Changes 👷‍♀️
- added `supportsAaveStopLoss` function
- used the above function when needed
- changed `AaveProtection` and `AaveProtectionWrite` flags so that it includes V3 (in its name)
  
## How to test 🧪
- try to open Aave V2 position, this should not support automation (no stop loss step before confirming)
- try to open Aave V3 position on mainnet, this should support automation (there is stop loss step before confirming)
- try to open Aave V3 position on optimism, this should not support automation (no stop loss step before confirming)


features/aave/helpers/supportsAaveStopLoss.ts
```diff
@@ -0,0 +1,9 @@
+import { NetworkIds } from 'blockchain/networks'
+import { LendingProtocol } from 'lendingProtocols'
+
+const supportedProtocols = [LendingProtocol.AaveV3]
+const supportedNetworks = [NetworkIds.MAINNET]
+
+export function supportsAaveStopLoss(protocol: LendingProtocol, network: NetworkIds) {
```
piotrwitek, line 7: as this is checking for V3 specifically, someone would make a bug using this with V2 with the expectation that this would work for V2 as the name deosn't imply it woudn't... perhaps renaming to `supportsAaveV3StopLoss` make sense? what you think?

features/automation/metadata/aave/stopLossMetadata.ts
```diff
@@ -31,188 +33,197 @@ import { prepareStopLossTriggerDataV2 } from 'features/automation/protection/sto
...
-    automationTriggersData,
-    triggerData: {
-      stopLossTriggerData: { isStopLossEnabled, stopLossLevel, triggerId, executionParams },
-    },
-    positionData: {
-      positionRatio,
-      liquidationRatio,
-      liquidationPrice,
-      liquidationPenalty,
+export function getAaveStopLossMetadata(lendingProtocol: LendingProtocol, networkId: NetworkIds) {
```
piotrwitek, line 41: this is not a getter function anymore, now it's a creator function that return a getter getAaveStopLossMetadata function.
IMO it should be rename to `createGetAaveStopLossMetadata` to make it clear to the maintainers

piotrwitek (commented):
@marcinciarka added some comments

features/aave/helpers/supportsAaveStopLoss.ts
marcinciarka: We want to check if a protocol/network pair is supported with stop loss. This list (both protocols and networks, hopefully) will grow.

features/automation/metadata/aave/stopLossMetadata.ts
marcinciarka: Good catch, fixed.

features/aave/helpers/supportsAaveStopLoss.ts
piotrwitek: ok make sense to me, thanks!

piotrwitek (approved):
LGTM

PR #2520, opened 2023-07-11
Create 1inch proxy, and route 1inch calls from front-end through it.

piotrwitek:
# [Create 1inch proxy, and route 1inch calls from front-end through it.](https://app.shortcut.com/oazo-apps/story/10212/create-1inch-proxy-and-route-1inch-calls-from-front-end-through-it)

## Changes 👷‍♀️
Implemented a new api route /api/exchange to proxy calls to the 1inch api with added authentication header.
Auth header is using a static key from .env which is put into the header under the `auth-key` prop.
Refactored swap helpers and exchange views to use a new api proxy.
  
## How to test 🧪
  Open eth multiply position which is using this proxy API to 1inch and follow up to complete the flow and open position succesfully. Check network tab in browser dev tools that it's using the proxy endpoint /api/exchange instead of 1inch api directly.


features/exchange/exchange.ts
```diff
@@ -11,7 +11,7 @@ import { Dictionary } from 'ts-essentials'
 
 import { amountFromWei, amountToWei } from '@oasisdex/utils/lib/src/utils'
 
-const API_ENDPOINT = `https://oasis.api.enterprise.1inch.exchange/v4.0/1/swap`
+const PROXY_API_ENDPOINT_SWAP = `/api/exchange/v4.0/1/swap`;
```
piotrwitek, line 14: I would like to extract `exchange` as a reusable route constant but I dunno where you store such things. Could you recommend something?

marcinciarka (approved):

PR #2521, opened 2023-07-11
🧿 Add balances to Ajna rewards page

piotrkonowrocki:
# [Add balances to Ajna rewards page](https://app.shortcut.com/oazo-apps/story/10016/remove-fortnightly-claim-period-add-balance-from-ajna-rewards)

image
  
## Changes 👷‍♀️

- Created `/api/ajna-rewards` endpoint for loading rewards data from database,
- created `getClaimedRewards` query for loading data from subgraph,
- created `useAjnaRewards` hook for handling rewards data,
- added allowing to disable claiming button on demand,
- removed "Earning on {{amount}} of positions".
  
## How to test 🧪

This has to be thoroughly tested when actual data is available. Currently, both database and subgraph response are empty, so some tweaking around formatting, precision etc. might be required.

### DIFF
--- a/features/ajna/rewards/useAjnaRewards.ts
+++ b/features/ajna/rewards/useAjnaRewards.ts
@@ -1,0 +1,85 @@
+import BigNumber from 'bignumber.js'
+import { useCustomNetworkParameter } from 'blockchain/networks'
+import { Rewards } from 'features/ajna/common/components/AjnaRewardCard'
+import { getAjnaRewardsData } from 'handlers/ajna-rewards/getAjnaRewardsData'
+import { useAccount } from 'helpers/useAccount'
+import { zero } from 'helpers/zero'
+import { useCallback, useEffect, useState } from 'react'
+
+interface AjnaRewardsParamsState {
+  rewards: Rewards
+  isError: boolean
+  isLoading: boolean
+  refetch(): void
+}
+
+const defaultRewards = {
+  tokens: zero,
+  usd: zero,
+}
+const errorState = {
+  rewards: defaultRewards,
+  isError: true,
+  isLoading: false,
+}
+
+export const useAjnaRewards = (): AjnaRewardsParamsState => {
+  const { walletAddress } = useAccount()
+  const [networkParameter] = useCustomNetworkParameter()
+  const [state, setState] = useState<AjnaRewardsParamsState>({
+    rewards: defaultRewards,
+    isError: false,
+    isLoading: true,
+    refetch: () => {},
+  })
+
+  const fetchData = useCallback(async (): Promise<void> => {
+    setState({
+      ...state,
+      isLoading: true,
+    })
+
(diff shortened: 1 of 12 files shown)


features/ajna/rewards/useAjnaRewards.ts
```diff
@@ -0,0 +1,87 @@
...
+            ReturnType<typeof getAjnaRewardsData>
+          >
+          const claimed = subgraphResponse.reduce((total, current) => total + current.amount, 0)
+
+          if (parseApiResponse.amount) {
+            setState({
+              ...state,
+              rewards: {
+                tokens: new BigNumber(parseApiResponse.amount).minus(claimed),
+                usd: zero,
```
piekczyk, line 59: please add here TODO saying that we should update it once Ajna token price will be available

features/ajna/rewards/useAjnaRewards.ts
```diff
@@ -0,0 +1,87 @@
...
+          })
+        })
+    }
+  }, [walletAddress, networkParameter?.id])
+
+  useEffect(() => void fetchData(), [fetchData])
+
+  return {
+    ...state,
+    refetch: useCallback(() => fetchData(), [fetchData]),
```
piekczyk, line 85: `refetch: fetchData` won't be enough since fetchData is already wrapped into `useCallback`? Second thing - `refetch` by its nature should be an async function, with this wrapper we are losing this information.

handlers/ajna-rewards/getAjnaRewardsData.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  const { address, networkId } = querySchema.parse(query)
+
+  try {
+    return {
+      amount: (
+        await prisma.ajnaRewardsWeeklyClaim.findMany({
+          where: {
+            user_address: address,
```
piekczyk, line 30: please cast it to lower case

piekczyk (changes requested):

features/ajna/rewards/useAjnaRewards.ts
piotrkonowrocki: Done.

features/ajna/rewards/useAjnaRewards.ts
piotrkonowrocki: Fixed and tested, I updated database in the meantime while page was already loaded and after triggering `refetch`, the data was updated.
Also, we don't need any information about if it's being async or anything else, because it's a standard async type use effect which contains additional metadata like `isLoading` or `isError`.

handlers/ajna-rewards/getAjnaRewardsData.ts
piotrkonowrocki: Fixed, but on the wrapping function level, not in a query.

piekczyk (approved):

PR #2532, opened 2023-07-12
TECH-1916 - Adding ONE_INCH_API_KEY env var to CICD pipeline and Dockerfile

OleksandrUA:
As requested in TECH-1916 adding ONE_INCH_API_KEY


piotrwitek (approved):
LGTM

PR #2570, opened 2023-07-18
:bug: user can't close postion from safe wallet

jakubswierczek:
Change first address in access list to user's proxy.


piotrwitek (approved):
Good

PR #2597, opened 2023-07-21
Ajna multiply price impact

piekczyk:
# Ajna multiply price impact

## Changes 👷‍♀️

- refactored `exchangeQuote$` observable to be able to use any token pairs
- added initial quote handling to Ajna multiply order information and based on that and value from swap calculated price impact
- adjusted slider logic to use prices from our price feed (maybe there is a way to base it on swap data, to be investigated) so slider works ok on short pools
- adjusted price impact label in order information to show values not in $ but in coll token / quote token
  
## How to test 🧪

- use positions on different pools for example ETH/USDC, USDC/ETH, CBETH/ETH and verify price impact
- on short position check whether slider works as expected

### DIFF
--- a/components/AppContext.ts
+++ b/components/AppContext.ts
@@ -807,7 +807,18 @@ export function setupAppContext() {
       amount: BigNumber,
       action: ExchangeAction,
       exchangeType: ExchangeType,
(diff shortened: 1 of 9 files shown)


BackToTheCode (approved):
All looks good to me. Just wanted to discuss

PR #2598, opened 2023-07-21
Mixpanel tracking for DSR

marcinciarka:
# [Mixpanel tracking for DSR](https://app.shortcut.com/oazo-apps/story/9539/mixpanel-tracking-for-dsr)
  
## Changes 👷‍♀️
- implemented mixpanel tracking for DSR input and depositing

### DIFF
--- a/analytics/analytics.ts
+++ b/analytics/analytics.ts
@@ -1166,7 +1166,8 @@ export const trackingEvents = {
         section: CommonAnalyticsSections.Form,
         id: {
           [EventTypes.InputChange]: 'DepositAmount',
-          [EventTypes.ButtonClick]: 'ConfirmDeposit',
+          [EventTypes.ButtonClick]:
+            eventData.action === 'deposit' ? 'ConfirmDeposit' : 'ConfirmWithdraw',
         }[event],
         page: Pages.DAISavingsRate,
         product: ProductType.EARN,
--- a/analytics/analytics.ts
+++ b/analytics/analytics.ts
@@ -1153,11 +1153,17 @@ export const trackingEvents = {
   },
   daiSavingsRate: (
     event: EventTypes.InputChange | EventTypes.ButtonClick,
-    eventData: { depositAmount: number; txHash?: string; network?: string; walletType?: string },
+    eventData: {
+      depositAmount: number
+      action: 'deposit' | 'withdraw'
+      txHash?: string
+      network?: string
+      walletType?: string
+    },
   ) => {
     !mixpanel.has_opted_out_tracking() &&
       mixpanelInternalAPI(event, {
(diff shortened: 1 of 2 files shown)


analytics/analytics.ts
```diff
@@ -179,6 +180,7 @@ export enum CommonAnalyticsSections {
   Form = 'Form',
   NotificationCenter = 'NotificationCenter',
   NotificationPreferences = 'NotificationPreferences',
+  OpenPosition = 'OpenPosition',
```
piekczyk, line 183: shouldn't we use here just `Form` as OpenPosition is rather a page-like thing?

features/dsr/helpers/dsrDeposit.tsx
```diff
@@ -210,6 +214,12 @@ function addTransitions(
   }
 
   if (state.stage === 'depositSuccess' || state.stage === 'withdrawSuccess') {
+    trackingEvents.daiSavingsRate(EventTypes.ButtonClick, {
```
piekczyk, line 217: I would distinguish between deposit and withdraw events using additional key in event payload, for example`action: 'deposit' | 'withdraw'`

features/dsr/utils/formUtils.tsx
```diff
@@ -1,13 +1,18 @@
+import { EventTypes, trackingEvents } from 'analytics/analytics'
 import BigNumber from 'bignumber.js'
 import { ManualChange } from 'features/dsr/helpers/dsrDeposit'
 
 export function handleAmountChange(change: (ch: ManualChange) => void) {
   return (e: React.ChangeEvent<HTMLInputElement>) => {
     const value = e.target.value.replace(/,/g, '')
+    const amount = value === '' ? undefined : new BigNumber(value)
+    if (amount) {
```
piekczyk, line 9: There was a requirement to track input? It may generate spam of events which IMO are meaningless until user actually send tx.

piekczyk (changes requested):

analytics/analytics.ts
marcinciarka: Changed that.

features/dsr/helpers/dsrDeposit.tsx
marcinciarka: Added that.

features/dsr/utils/formUtils.tsx
marcinciarka: Product team wants it (doesn't see this as an issue).

piekczyk (dismissed):

piekczyk (approved):

jakubswierczek (approved):

PR #2612, opened 2023-07-25
Show Fee Split inline (Ajna Multiply)

BackToTheCode:
Screenshot 2023-07-25 at 12 53 22

  
## Changes 👷‍♀️
- Show t/x fee and gas fee as separate items in transaction fee row in order info
- Hide oasis fee portion when zero (for example earn multiply)
- Hide dropdown when oasis fee portion is zero
  
## How to test 🧪
- Set up Ajna multiply position
- Observe whether t/x is broken apart


features/ajna/positions/multiply/sidebars/AjnaMultiplyFormOrder.tsx
```diff
@@ -237,8 +238,14 @@ export function AjnaMultiplyFormOrder({ cached = false }: { cached?: boolean })
...
           ? [
               {
-                label: t('system.max-transaction-cost'),
-                value: <GasEstimation addition={oasisFee} />,
+                label: t('transaction-fee'),
+                value: (
+                  <>
+                    {formatted.oasisFee}
+                    <Box sx={{ mx: '4px' }}>+</Box>
+                    <GasEstimation addition={oasisFee} />
```
piekczyk, line 246: I didn't test it but it seems that it will be displayed for borrow actions as well. We should add here condition whether `oasisFee` is zero and display it only if above zero

piekczyk (approved):

PR #2627, opened 2023-07-27
Ajna handle multiple positions for same proxy

piekczyk:
# [Ajna handle multiple positions for same proxy](https://app.shortcut.com/oazo-apps/story/10643/ui-provide-support-for-many-ajna-positions-on-the-same-proxy)

## Changes 👷‍♀️

- added handling for extra parameters in URL beside position ID
- added support for many positions on the same proxy for Ajna
  
## How to test 🧪

- use owner page of following user `0xd790A1516F78E3FB52338084B5C5931a75BB19CF` and try to visit both `963` positions


features/vaultsOverview/parsers.tsx
```diff
@@ -258,7 +258,7 @@ export function parseAjnaBorrowPositionRows(
         isAtRiskWarning: false,
         type: 'LTV',
       },
-      url: `/ethereum/ajna/${vaultId}`,
+      url: `/ethereum/ajna/borrow/${collateralToken}-${quoteToken}/${vaultId}`,
```
jakubswierczek, line 261: I thought the `vaultId` would be first after the network and protocol.

piotrkonowrocki: Sorry for misleading, that was my idea to do that in that way in the end. This is more coherent,
`/network/protocol/product/collateral/quote` - open flow
`/network/protocol/product/collateral/quote/id` - but just add id at the end, and it becomes manage flow.

piotrkonowrocki (approved):

jakubswierczek (approved):

PR #2630, opened 2023-07-27
chore: [sc-9982] Separate AppContext from the static pages

marcinciarka:
Story details: https://app.shortcut.com/oazo-apps/story/9982


components/layouts/AppLayout.tsx
```diff
@@ -0,0 +1,25 @@
+import { Footer } from 'components/Footer'
```
piotrwitek, line 1: @marcinciarka I really like breaking these silos down into smaller files 👍

piotrwitek: btw. you could add an index file here reexporting all Layout components from index. It would do two things:
1) no need to refactor old imports
2) I think that's easier to import from jsut 'components/layouts' and it's easier to refactor/rename components without breaking consumers as refactoring works much better.
What you think @marcinciarka?

piotrkonowrocki (approved):

components/context/MainContextProvider.tsx
```diff
@@ -0,0 +1,27 @@
+import { MainContext, setupMainContext } from 'helpers/context/MainContext'
```
jakubswierczek, line 1: Since we have a lot of contexts right now, we should create an `index.ts` file to export them. 
What about our naming conventions?

components/context/AccountContextProvider.tsx
```diff
@@ -0,0 +1,357 @@
...
+      const urnResolver$ = curry(createVaultResolver$)(
+        cdpManagerIlks$,
+        cdpManagerUrns$,
+        charterUrnProxy$,
+        cropperUrnProxy$,
+        cdpRegistryOwns$,
+        cdpManagerOwner$,
+        proxyOwner$,
+      )
+      const charterCdps$ = memoize(
```
jakubswierczek, line 188: Is that related to an account? 
I think it should be placed in the "Maker" specific context.

helpers/bigNumberToString.ts
```diff
@@ -0,0 +1,5 @@
+import BigNumber from 'bignumber.js'
```
jakubswierczek, line 1: Do we really need this function?

helpers/context/MainContext.ts
```diff
@@ -0,0 +1,75 @@
+import { createSend } from '@oasisdex/transactions'
```
jakubswierczek, line 1: I would prefer to have it in one "context" folder. Right now, the helpers are in a separate module, but they should be placed together with their usage. So, I will have a module called "context" which will contain all the necessary functions/components.

helpers/context/MainContext.ts
```diff
@@ -0,0 +1,75 @@
...
+import { createGasPrice$ } from 'blockchain/prices'
+import { createWeb3Context$ } from 'features/web3Context'
+import { createTxHelpers$ } from 'helpers/createTxHelpers'
+import { Observable, of } from 'rxjs'
+import { distinctUntilChanged, shareReplay, switchMap } from 'rxjs/operators'
+
+import { DepreciatedServices, TxData, TxHelpers$ } from './types'
+
+export function setupMainContext() {
+  console.log('Main context setup')
```
jakubswierczek, line 22: Remove log.

package.json
```diff
@@ -19,6 +19,7 @@
     "lint": "eslint --ext .ts --ext .tsx .",
     "lint:fix": "yarn lint --fix",
     "typecheck": "tsc -p ./tsconfig.test.json --noEmit && tsc -p ./tsconfig.json --noEmit",
+    "typecheck-watch": "tsc -p ./tsconfig.json --noEmit --watch",
```
jakubswierczek, line 22: `typecheck:watch` will fit better into our script convention.

jakubswierczek (changes requested):
Summary of pull request review comments:

- Create an `index.ts` file to export the contexts.
- Consider placing the "Maker" specific context in the "Account" context folder.
- Add an `index.ts` file to the `AppLayout` component folder.
- Evaluate the necessity of the `bigNumberToString` function.
- Consolidate the helpers and their usage in one "context" folder.
- Remove the log statement in the `MainContext` helper.

components/context/MainContextProvider.tsx
marcinciarka: Done.

jakubswierczek (approved):

PR #2642, opened 2023-07-31
Identify tokens

piekczyk:
# [Identify tokens](https://app.shortcut.com/oazo-apps/story/10588/identified-tokens-stored-in-database)

## Changes 👷‍♀️

- adjusted `idenftifyTokens` observable
- added caching layer in db
  
## How to test 🧪

- use `identifyTokens$` in any place you want with tokens addresses, verify whether tokens not available in local config has been saved to db to avoid unnecesary api / rpc calls


handlers/risk/tokens.ts
piotrkonowrocki, line 1: Why is this file in `risk` directory?

handlers/tokens-info/get.ts
```diff
@@ -0,0 +1,72 @@
...
+  let tokensFromApi: Tokens[] = []
+  const addressesFromDb = tokensFromDb.map((token) => token.address)
+  try {
+    tokensFromApi = await readTokensFromApi(
+      lowerCasedAddresses.filter((address) => !addressesFromDb.includes(address)),
+    )
+
+    await saveTokensToDb(tokensFromApi)
+
+    if (tokensFromApi.length === lowerCasedAddresses.length) {
```
piotrkonowrocki, line 47: That part should also verify if tokens identified by API and tokens identified by DB are already full collection, because maybe there's no need to go to next step.

I think instead of stuff like
```
if (tokensFromDb.length === lowerCasedAddresses.length) {
```
and
```
if (tokensFromApi.length === lowerCasedAddresses.length) {
```
there could be a single `const` that is updated after every step. If amount of items in that array is enough, skip rest of the checks. But that would probably required changing array to object with unique keys, because there is a possibility that one token might be identified by both db and api, but the other has to be loaded from blockchain. But it won't since length of global array is already 2.

blockchain/identifyTokens.ts
```diff
@@ -1,77 +1,59 @@
...
+          precision: getToken(token).precision,
+          address: tokensContracts[token].address,
+          chain_id: context.chainId,
+        }))
+
+        if (tokensAddresses.length === localTokensAddresses.length) return of(identifiedTokens)
+      }
+
+      return from(
+        fetch(`/api/tokens-info`, {
```
piotrkonowrocki, line 42: Now we have `/api/tokens-info` and `/api/token-identifier`. Maybe rename my API endpoint, so they don't look like they are doing the same thing?

piotrkonowrocki (changes requested):

piotrkonowrocki (approved):

PR #2677, opened 2023-08-07
[Network Switcher] Goerli Support

jakubswierczek:
# Changes
- I removed the `NetworkConfig` from local storage, but I had to add a new value: `legacy-default-chain`. This value is used as the default chain ID for the network connector. I needed to set it up because some of the pipelines from Maker use the default chain ID from the first render, which was causing issues: 
> Error: Returned values aren't valid, did it run Out of Gas? You might also see this error if you are not using the correct ABI for the contract you are retrieving data from, requesting data from a block number that does not exist, or querying a node which is not fully synced.,

The refresh is performed when switching between Ethereum and Goerli.


blockchain/ens.ts
```diff
@@ -3,14 +3,12 @@ import { Observable } from 'rxjs'
...
-import { getNetworkRpcEndpoint, NetworkIds } from './networks'
+import { getNetworkRpcEndpoint } from './networks'
 
 export function resolveENSName$(context$: Observable<Context>, address: string) {
   return context$.pipe(
     switchMap(async (context) => {
-      const provider = new ethers.providers.JsonRpcProvider(
-        getNetworkRpcEndpoint(NetworkIds.MAINNET, context.chainId),
-      )
+      const provider = new ethers.providers.JsonRpcProvider(getNetworkRpcEndpoint(context.chainId))
```
piotrwitek, line 11: Hey @jakubswierczek `getNetworkRpcEndpoint` is accepting NetworkId, not chainId, it looks confusing, we should refactor that:
```export function getNetworkRpcEndpoint(networkId: NetworkIds) {```

Comment (piotrwitek) [reply blockchain/ens.ts:11]: We could also consider doing opposite way, renaming NetworkId type to ChainId, what would be better?

File: features/web3OnBoard/use-safafty-reload.tsx [hunk 2]
```diff
@@ -0,0 +1,32 @@
...
+import { useLegacyDefaultChain } from './use-legacy-default-chain'
+import { WalletManagementState } from './wallet-state'
+
+/*
+ * I don't like this solution,
+ * but it's the only way I found to make it work.
+ * The problem is that some components strictly depend on Mainnet or Goerli,
+ * so the only way to refresh them is to reload the page.
+ */
+export function useSafaftyReload({ walletNetworkHexId }: WalletManagementState) {
```
piotrwitek, line 15: what safafty means? it's not a word

piotrwitek (changes requested):
Just a typo

piotrwitek (approved):

PR #2700, opened 2023-08-10
Show price ratio in aave headers

jakubswierczek:
# Show price ratio in aave header

## Changes
- On aave multiply, borrow and manage earn instead of collateral price and debt price we show ratio between those prices. In case of short position we show debt to collateral. 

## How to test
- Go to open aave postion
- Go to manage aave positon.


BackToTheCode (dismissed):
Approved pending typecheck issue.

BackToTheCode (approved):

PR #2717, opened 2023-08-16
:bug: Fix infinite reloading of Aave Header

jakubswierczek:
# Changes
-  Remove bug in Aave header that causes infinite reloading.
- The button `Go to position` works right now.

Story details: https://app.shortcut.com/oazo-apps/story/10904

### DIFF
--- a/features/aave/components/AaveHeader.tsx
+++ b/features/aave/components/AaveHeader.tsx
@@ -1,7 +1,6 @@
 import { Protocol } from '@prisma/client'
 import { ProtocolLabelProps } from 'components/ProtocolLabel'
 import { VaultHeadline } from 'components/vault/VaultHeadline'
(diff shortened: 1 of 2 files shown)


features/aave/components/AaveHeader.tsx
```diff
@@ -29,42 +27,32 @@ function AaveHeader({
...
-      }`,
-    })
-  }
+    return [
+      {
+        label: t('aave.header.current-market-price'),
+        value: `${formatCryptoBalance(positionPrice)} ${priceFormat}`,
+      },
+    ]
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
BackToTheCode, line 54: Could we just include function `t` in the dependencies array and not worry about disabling the rule?

BackToTheCode (approved):

PR #2735, opened 2023-08-18
Liquidation Price & Net Value - Aave Positions

jakubswierczek:
# Changes
- Refactor way of displaying Liquidation Price and Net Value of the position in Aave
- New components only for `NetValue` & `LiquidationPrice` 
- In case of short position tyle we are displaying Debt/Collateral instead of Collateral/Debt
- It's a just view change without changing any significant calculations.


BackToTheCode (approved):
Lovely

PR #2736, opened 2023-08-18
feat: [sc-11043] [Referrals][frontend] - ENS name stretch goal

marcinciarka:
Story details: https://app.shortcut.com/oazo-apps/story/11043


blockchain/ens.ts
```diff
@@ -17,3 +20,69 @@ export function resolveENSName$(context$: Observable<Context>, address: string)
     }),
   )
 }
+
+// address => ens name
+export async function addressToEnsNameMainnet(ensName: string) {
+  const provider = new ethers.providers.JsonRpcBatchProvider(
+    getNetworkRpcEndpoint(NetworkIds.MAINNET),
+  )
+  return await provider.lookupAddress(ensName).catch((err: Error) => {
```
piotrwitek, line 29: I would suggest to clean those awaits from the return statements ,async functions always returns promises So those awaits are redundant

blockchain/ens.ts
```diff
@@ -17,3 +20,69 @@ export function resolveENSName$(context$: Observable<Context>, address: string)
     }),
   )
 }
+
+// address => ens name
+export async function addressToEnsNameMainnet(ensName: string) {
```
piotrwitek, line 25: I'm really confused what this function does, it accept ens name as argument but it's supposed to map address to the ens name also The comment above suggest it accept address and return ens name, that means the argument name is Incorrect

blockchain/ens.ts
```diff
@@ -17,3 +20,69 @@ export function resolveENSName$(context$: Observable<Context>, address: string)
...
+    getNetworkRpcEndpoint(NetworkIds.MAINNET),
+  )
+  return await provider.lookupAddress(ensName).catch((err: Error) => {
+    console.warn(`Error looking up ENS name for address: ${ensName}/${err.message}`)
+    return null
+  })
+}
+
+// ens name => address
+export async function ensNameToAddressMainnet(address: string) {
```
piotrwitek, line 36: here is the same issue as above regarding the argument name

blockchain/ens.ts
```diff
@@ -17,3 +20,69 @@ export function resolveENSName$(context$: Observable<Context>, address: string)
...
+  const provider = new ethers.providers.JsonRpcBatchProvider(
+    getNetworkRpcEndpoint(NetworkIds.MAINNET),
+  )
+  return await provider.resolveName(address).catch((err: Error) => {
```
piotrwitek, line 40: remove await

blockchain/ens.ts
```diff
@@ -17,3 +20,69 @@ export function resolveENSName$(context$: Observable<Context>, address: string)
...
+      .catch((err: Error) => {
+        console.warn(`Error looking up ENS name for address: ${err.message}`)
+      })
+  }, [address])
+
+  return [ensName]
+}
+
+export const useMainnetEnsNames = (addresses?: string[]) => {
+  const firstAddressesList: { [key: string]: string } = {}
```
piotrwitek, line 64: initialAddressesList

features/referralOverview/FeesView.tsx
```diff
@@ -22,7 +24,8 @@ export interface UpsertUser {
 
 export function FeesView({ userReferral }: Props) {
   const { t } = useTranslation()
-
+  const [ensReferrerName] = useMainnetEnsName(userReferral?.referrer)
+  const ensReferredNames = useMainnetEnsNames(userReferral.referrals)
```
piotrwitek, line 28: These api return values are not particularly consistent, first one is wrapped in an array and the other one is not. I would Suggest to strive for consistent Behavior For reusable modules as it is much easier to use those

piotrwitek (changes requested):

blockchain/ens.ts
marcinciarka: Good catch.

blockchain/ens.ts
marcinciarka: 👌

blockchain/ens.ts
marcinciarka: 👌

blockchain/ens.ts
marcinciarka: 👌

features/referralOverview/FeesView.tsx
marcinciarka: fixed

piotrwitek (approved):

PR #2739, opened 2023-08-18
fix: [sc-11176] [BUG][Aave v3][Borrow] UI issue in Management block when opening an existent position's page

jakubswierczek:
Story details: https://app.shortcut.com/oazo-apps/story/11176

1. Back to editing button - only visible when reviewing adjusting
2. Default manage collateral operation set.
3. Disable button when no amount set.


BackToTheCode (approved):
Approved. `Back to edit` text link only visible to user when switching position type. Confirming that as intended behaviour. Code is fab

PR #2808, opened 2023-08-29
[sc-11316] [sc-11317] Extends Ajna borrow with close feature 

paszkowskiDamian:
# [Close and Adjust action on ajna borrow](https://shortcutLink)
  
## Changes 👷‍♀️
- Adds cose and adjsut functionality to ajna borrow
  
## How to test 🧪
- Open or navigate to borrow position
- The dropdown should contain two new entries: `borrow` and `adjust`
- Select `adjust`
- Form should allow you to adjust your position, both ways to increase and decrease risk, the order confirmation should contain relevant data for borrow position (it will be different to multiply)
- Test if transaction passes and the resulting position is as expected 
- Select `close` from dropdown
- Now form should allow you to close position to either collateral or debt token
- Try if transaction passes

Gotchas:
- The two additional functionalities shouldn't appear for pools where multiply is not supported, e.g. custom pools


actions/ajna/multiply.ts
```diff
@@ -84,7 +85,7 @@ export const ajnaOpenMultiply = ({
   )
 }
 
-export const ajnaAdjustMultiply = ({
+export const ajnaAdjust = ({
```
piotrkonowrocki, line 88: Since this is no longer multiply action, is should be moved into `actions/ajna/common.ts` file (which does not exist right now.

actions/ajna/multiply.ts
```diff
@@ -133,7 +134,7 @@ export const ajnaAdjustMultiply = ({
   )
 }
 
-export const ajnaCloseMultiply = ({
+export const ajnaClose = ({
```
piotrkonowrocki, line 137: Same as with `ajnaAdjust`.

features/ajna/positions/borrow/sidebars/AjnaBorrowFormContentClose.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+import { AjnaBorrowFormOrder } from './AjnaBorrowFormOrder'
+
+export function AjnaBorrowFormContentClose() {
+  const { t } = useTranslation()
+  const {
+    environment: { collateralToken, collateralPrice, quoteToken, quotePrice, product },
+  } = useAjnaGeneralContext()
+
+  if (product === 'earn') {
```
piotrkonowrocki, line 19: I think this condition is no longer needed and it's probably leftover from early development stages? Since `useAjnaProductContext('borrow')` is pointing directly to `borrow` product type, `earn` will never show in here.

features/ajna/positions/common/components/AjnaAdjustSlider.tsx
```diff
@@ -16,26 +16,31 @@ const min = new BigNumber(0.01)
...
+export function AjnaAdjustSlider({ disabled = false }: AjnaAdjustSliderProps) {
   const { t } = useTranslation()
   const {
-    environment: { collateralToken, quoteToken, isShort },
+    environment: { collateralToken, quoteToken, isShort, product },
     steps: { currentStep },
   } = useAjnaGeneralContext()
+
+  if (product === 'earn') {
+    return null
```
piotrkonowrocki, line 31: How about throwing an error similar to what we have in `useAjnaProductContext`, instead of returning `null`? Might be easier to debug with `throw new Error`.

piotrkonowrocki (changes requested):

piotrkonowrocki (approved):

piotrwitek (approved):

PR #2824, opened 2023-09-01
feat: [sc-11192] ✨ Spark Product Hub filters + select filters feature flag support

marcinciarka:
Added Spark protocol to product hub filters
Added filter options feature flag support

### DIFF
--- a/features/productHub/meta.ts
+++ b/features/productHub/meta.ts
@@ -1,5 +1,6 @@
 import { BaseNetworkNames, NetworkNames, networksByName } from 'blockchain/networks'
 import { getToken } from 'blockchain/tokensMetadata'
+import { GenericMultiselectOption } from 'components/GenericMultiselect'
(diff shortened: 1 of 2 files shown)


features/productHub/meta.ts
```diff
@@ -208,5 +209,12 @@ export const productHubProtocolFilter = [
     label: lendingProtocolsByName[LendingProtocol.Ajna].label,
     value: lendingProtocolsByName[LendingProtocol.Ajna].name,
     image: lendingProtocolsByName[LendingProtocol.Ajna].icon,
+    featureFlag: 'AjnaSafetySwitch',
```
piotrkonowrocki, line 212: Ajna landing page sets Ajna as protocol by default in filters, have you checked what's going to happen if you open that page?

marcinciarka: Removed it as it is backwards compared to the standard feature flags (default on vs default off).

piotrkonowrocki (approved):

PR #2825, opened 2023-09-01
[sc-11627] Add banner for Ajna borrow positions 

paszkowskiDamian:
# Add banner for Ajna positions
  
## Changes 👷‍♀️
- Supported borrow ajna positions should display a banner with info about new features.
  
## How to test 🧪
- Go to ajna position, 

The link to the documentation will be provided next week.

### DIFF
--- a/features/ajna/positions/common/notifications.tsx
+++ b/features/ajna/positions/common/notifications.tsx
@@ -29,6 +29,7 @@ import { EXTERNAL_LINKS } from 'helpers/applicationLinks'
 import { one, zero } from 'helpers/zero'
 import { Trans } from 'next-i18next'
 import React, { Dispatch, FC } from 'react'
+import { isPoolSupportingMultiply } from './helpers/isPoolSupportingMultiply'
 
 interface AjnaLiquidationNotificationWithLinkProps {
   translationKey: string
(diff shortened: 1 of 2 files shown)


public/locales/en/common.json
```diff
@@ -2692,6 +2692,10 @@
           "cta-textual": "Open new position"
         },
         "notifications": {
+          "risk-managemnt": {
```
piotrwitek, line 2695: typo in managemnt

piotrwitek (changes requested):
checks are failing

features/ajna/positions/common/notifications.tsx
```diff
@@ -368,6 +387,11 @@ export function getAjnaNotifications({
         notifications.push(ajnaNotifications.aboveLup(null))
       }
 
+      if (isPoolSupportingMultiply({collateralToken, quoteToken})) {
+        notifications.push(ajnaNotifications.multiplyNotice(null))
+      }
```
piotrwitek, line 392: I'm trying to understand this code when reading it and to me it seems like It is showing a multiply notice for the pools supporting the multiply, But from the requirements it was supposed to display risk management notification for borrow positions. Why the naming convention is so different?

PR #2836, opened 2023-09-05
chore: [sc-11708] CodeSandbox - Usability within the front-end app.

jakubswierczek:
# Changes 
- The `.codesandbox` directory contains the `Dockerfile`, which serves as the base environment for the Code Sandbox MicroVM. In the `docker-compose` file, I added Postgres and Pgadmin to enable data viewing from the database. The `Tasks.json` file includes definitions for setup tasks such as package installation or migration. Upon machine startup, the production build is initiated.
- Thanks to the directory, we have the deployment link.: 
CleanShot 2023-09-08 at 19 37 54@2x

- Removed `API_HOST` and `APP_FULL_DOMAIN`. Right now, all addresses are relative, so we can easily change the domain for our app. That change requires modifying the behavior in `sdaiOracle` and updating `product-hub`, as there were server actions. I changed the requests to itself to execute the correct code.
- Removed `tsconfig.dev.json`. The migration and new script (`update-product-hub`) are now started with `npx ts-node`.
- Removed the `.env` file from the repository and introduced a `.env.template` file. I needed to remove it because the preview environment retrieves data from that file, and it hides virtual machine environment settings.
- Removed linting from `next build`. 
- Removed `NODE_ENV` from `next build`


marcinciarka (approved):

scripts/update-product-hub.ts
```diff
@@ -0,0 +1,21 @@
+import dotenv from 'dotenv'
+import fetch from 'node-fetch'
+dotenv.config()
+
+export const updateProductHub = async () => {
+  const result = await fetch('http://localhost:3000/api/product-hub', {
+    method: 'PATCH',
+    body: JSON.stringify({
+      protocols: ['ajna', 'aavev2', 'aavev3', 'maker', 'sparkv3'],
```
piotrwitek, line 9: @jakubswierczek I would suggest to update the endpoint to use all available protocols by default when protocols argument is undefined. I think this is main use case anyway so would be great to have such default behavior.

Also this would make this module work for all protocols added in the future. Otherwise we'll need to update this code each time we add a new protocol.

PR #2842, opened 2023-09-06
Restore borrow UI for multiply

paszkowskiDamian:
# Restore borrow UI 
  
## Changes 👷‍♀️
- Removes all unnecessary files
- Restores Borrow forms 
  
## How to test 🧪
- Open borrow position
- Now you should be able to borrow and deposit more, or payback dai and withdraw collateral 
- when switched to multiply, you can do only one operation but you may adjust your position (withdraw&adjust etc)


blockchain/calls/proxyActions/proxyActions.ts
```diff
@@ -467,6 +467,22 @@ export const closeVaultCall: TransactionDef<CloseVaultData> = {
...
+    data.exchangeData,
+    
+    
+    
+    
+    
+    
+    
+    ${data.exchangeData}`)
+
```
piotrwitek, line 485: cleanup

PR #2854, opened 2023-09-07
[fix] Ensure ETH can token symbol be sent to library

BackToTheCode:
* Fixes issue where FE was mapping position symbols to WETH because of library type
* Library needs ETH as symbol in order to determine whether or not to skip wrap/unwrap WETH actions in operations


marcinciarka (approved):

marcinciarka (changes requested):
The `getCurrentPositionLibCallData` is used in `AaveBorrowPositionData` and `AaveMultiplyPositionData` to display the correct token.

marcinciarka (approved):

PR #2864, opened 2023-09-11
[sc-12017] - RPC Gateway

halaprix:

  
## Changes 👷‍♀️
- added `rpcGateway` API endpoint that call RPC Gateway instead of provider directly https://github.com/OasisDEX/rpc-gateway
- added `UseRpcGateway` feature flag
- added `RPC_GATEWAY` env
- added method to fetch the config and cache for backend with default TTL of 5 minutes
  
## How to test 🧪
- set `UseRpcGateway` to `true` locally
- set `RPC_GATEWAY` to https://discord.com/channels/837076147694207067/1156237535689850990/1156239716505620510
- see if it works :1st_place_medal:


config/rpcConfig.ts
```diff
@@ -1,14 +1,23 @@
 import { NetworkNames } from 'blockchain/networks/network-names'
 import { clientId } from 'helpers/clientId'
+import { getLocalAppConfig } from 'helpers/config'
 
 import { infuraProjectId } from './runtimeConfig'
 
+const { UseRpcGateway } = getLocalAppConfig('features')
```
marcinciarka, line 7: Consider moving that inside the function. It will not update with a config change now I think.

marcinciarka (approved):

PR #2865, opened 2023-09-11
🔧 [AWS JSON config] Implement the new JSON configs in oasis-borrow

marcinciarka:
## Dynamic App Config

Got rid of the in-app feature toggles. Now they are dynamically manages by a JSON file generated in https://github.com/OasisDEX/oazo-configuration.

## Changes:
- **new endpoint** - `/api/config` pulls the JSON file and caches it for 2 minutes
- **new context** - `ConfigContext`, takes care of pulling the config from `/api/config` (every 2 minutes) and providing that to `getAppConfig` method. This also updates the async config saved in the localstorage. That was necessary to allow access to the config everywhere (e.g. fork-settings, which is a method outside of a context scope).
- `Feature` is now an enum
- `getFeatureToggle`is no more, use `getAppConfig` if in context and `getAppConfigSync`if outside of context


helpers/config/access-config-context.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+}
+
+/**
+ * Returns currently saved config from localStorage
+ * PLEASE NOTE THAT THIS IS NOT DYNAMIC, IT WILL NOT UPDATE WHEN CONFIG CHANGES (only after a refresh)
+ * @param configKey
+ * @returns ConfigResponseType[T] or empty config
+ */
+export function getAppConfigSync<T extends ConfigResponseTypeKey>(
```
piekczyk, line 64: maybe `getLocalAppConfig`? When I first saw this method in code I thought it's for forcing re-sync of config from AWS

piekczyk (changes requested):

components/context/ConfigContextProvider.tsx
```diff
@@ -0,0 +1,45 @@
...
+          setContext({ config })
+          saveConfigToLocalStorage(config)
+        } catch (error) {
+          console.error(error)
+        }
+      }
+      await fetchConfig()
+      setInterval(fetchConfig, 1000 * configCacheTime.frontend)
+    }
+    void setup()
```
jakubswierczek, line 42: please add here `catch`

components/context/ConfigContextProvider.tsx
```diff
@@ -0,0 +1,45 @@
...
+  ConfigResponseType,
+  Feature,
+  saveConfigToLocalStorage,
+} from 'helpers/config'
+import { WithChildren } from 'helpers/types'
+import React, { useEffect, useState } from 'react'
+
+const configFetcher = () =>
+  fetch(`/api/config`, {
+    method: 'POST',
```
jakubswierczek, line 13: why `POST` ? we don't save anything, we just want to obtain the data. We should use `GET`

handlers/config/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  const response = await fetch(configUrl)
+  const data = await response.json()
+  return data
+}
+
+/**
+ * Fetches config from backend, caches it for configCacheTime and returns it
+ * @returns Record<string, string>
+ */
+export const configHandler = cacheObject<Record<string, string>>(
```
jakubswierczek, line 21: I don't know if we want to have cache for it. 
I think in first iteration we can use it without caching.

pages/api/config.ts
```diff
@@ -0,0 +1,11 @@
+import { configHandler } from 'handlers/config'
+import { NextApiRequest, NextApiResponse } from 'next'
+
+export default async function (req: NextApiRequest, res: NextApiResponse) {
+  const config = await configHandler()
+  const { method } = req
+  if (method === 'POST') {
```
jakubswierczek, line 7: please use `GET` and cache header without caching in memory

jakubswierczek (changes requested):

jakubswierczek (approved):

piekczyk (dismissed):

features/borrow/open/tests/openVault.test.ts
```diff
@@ -15,6 +17,12 @@ import { map } from 'rxjs/operators'
 
 import { newCDPTxReceipt } from './fixtures/newCDPtxReceipt'
 
+jest.mock('helpers/config', () => ({
```
jakubswierczek, line 20: ❤️

jakubswierczek (approved):

piekczyk (approved):

jakubswierczek (approved):

PR #2874, opened 2023-09-14
Enable Borrow on L2s

jakubswierczek:
@ Changes
- Enable open borrow position on L2
- Enable feature flags for borrow.


actions/aave-like/open/get-open-deposit-borrow-position-parameters.ts
```diff
@@ -28,8 +22,6 @@ export async function getOpenDepositBorrowPositionParameters(
     protocol,
   } = args
 
-  assertNetwork(networkId)
```
robercano, line 31: Removing this means that we are also supporting Polygon. Is this the case?

robercano (changes requested):

actions/aave-like/open/get-open-deposit-borrow-position-parameters.ts
jakubswierczek: No, Polygon is not supported in our app. We have a configuration, but we don't have Aave implemented for it.

robercano: Discussed offline and the library already have this check. This way we support borrow for all networks, and the library handles the error.

robercano (approved):

PR #2892, opened 2023-09-18
Package updates + lifi widget config update

marcinciarka:
# Package chores

  
## Changes 👷‍♀️
- Updated next
- removed @next/font
- updated web3-onboard
- ramda has to be transpiled
- updated lifi widget and [its config](https://app.shortcut.com/oazo-apps/story/11898/update-config-for-lifi-swap) 
- node 18 yay

## How to test 🧪
- app should work as before (a bit faster?)


jakubswierczek (dismissed):

test-configs/setup.ts
```diff
@@ -1,4 +1,5 @@
 import '@testing-library/jest-dom'
+import '@inrupt/jest-jsdom-polyfills'
```
piotrwitek, line 2: Could you leave a comment please for history why this was added?

piotrwitek (changes requested):

test-configs/setup.ts
marcinciarka: @jakubswierczek ⬆️

jakubswierczek: Some libraries are using certain parts of the Web API that do not appear in jsdom (which is used in the Jest framework). 

Error:
```
ReferenceError: TextEncoder is not defined
```

As a result, some of the tests are skipped, but we can enable that now (in the follow up). 

Polyfills can resolve this issue.

jakubswierczek (approved):

piotrwitek (approved):
looks great 🐸

PR #2895, opened 2023-09-20
Generate AppConfig type on app launch and build

marcinciarka:
# Generate AppConfig type on app launch and build

  
## Changes 👷‍♀️
- added `get-config-types` script
- it pulls the config from `CONFIG_URL` env and creates `types/config` type with the current config and feature enums

### DIFF
--- a/scripts/get-config-types.ts
+++ b/scripts/get-config-types.ts
@@ -1,0 +1,47 @@
+import dotenv from 'dotenv'
+import { writeFile } from 'fs/promises'
+import JsonToTS from 'json-to-ts'
+import fetch from 'node-fetch'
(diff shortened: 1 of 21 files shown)


piotrwitek (approved):

scripts/get-config-types.ts
```diff
@@ -0,0 +1,70 @@
...
+dotenv.config({
+  path: '.env',
+})
+dotenv.config({
+  path: '.env.local',
+})
+
+const getConfig = async () => {
+  const response = await fetch(process.env.CONFIG_URL as string)
+  return await response.json()
```
piotrwitek, line 15: await can be disposed

PR #2950, opened 2023-10-09
Product Finder filters as query string

piotrkonowrocki:
# Product Finder filters as query string

Partially covers https://app.shortcut.com/oazo-apps/story/12030/default-selection-of-protocol-in-product-finder
  
## Changes 👷‍♀️

- Added support for persistent filter settings kept between different product and tokens selected in Product Finder's natural language selector.
- Added support for initial load of filters based on query string state.
  
## How to test 🧪

**Important note:** use built version of application, not developer mode. As developer mode rebuilds each page on demand, sometimes it leads to unexpected results with content updating based on URL.

Please test following scenarios:
- Are all types of filters kept in the query string?
- Are filters persistent when natural language is updated?
- Are filters specific for only one type of product (e.g. debt token for Borrow) removed from query string when product type is changed?
- Are filters prefilled with options when page is open with query string?


features/productHub/views/ProductHubView.tsx
```diff
@@ -84,11 +93,33 @@ export const ProductHubView: FC<ProductHubViewProps> = ({
...
+              getInitialFilters({
+                initialQueryString,
+                initialNetwork,
+                initialProtocol,
+                selectedProduct: _selectedProduct,
+                token,
+              }),
+            )
+            setQueryString(
+              Object.keys(queryString).reduce<ProductHubQueryString>(
```
piekczyk, line 109: maybe we could extract it as method with meaningful name

piekczyk (changes requested):

features/productHub/views/ProductHubView.tsx
piotrkonowrocki: Good point, done.

piekczyk (approved):

PR #2960, opened 2023-10-11
Protocol general context

piekczyk:
# Protocol general context

## Changes 👷‍♀️

- Generalized AjnaGeneralContext
  
## How to test 🧪
- step 1 ...


features/ajna/positions/common/getAjnaSidebarTitle.tsx
```diff
@@ -11,10 +11,10 @@ export const getAjnaSidebarTitle = ({
   position,
   isOracless,
 }: {
-  currentStep: AjnaSidebarStep
-  product: AjnaProduct
+  currentStep: ProtocolSidebarStep
+  product: ProtocolProduct
   isFormFrozen: boolean
-  position: AjnaPosition | AjnaEarnPosition
+  position: BorrowishPosition | AjnaEarnPosition
```
piotrwitek, line 17: what is BorrowishPosition?

actions/ajna/oasisActionsLibWrapper.ts
```diff
@@ -32,8 +34,8 @@ interface AjnaTxHandlerInput {
   collateralToken: string
   context: Context
   isFormValid: boolean
-  position: AjnaGenericPosition
-  simulation?: AjnaGenericPosition
+  position: AjnaPosition | AjnaEarnPosition
+  simulation?: AjnaPosition | AjnaEarnPosition
```
piotrwitek, line 38: Isn't it better to extract this type as union type and add it to the library? I see that in a lot of places we have issues with adding new variation mainly because poor sharing, then weneed to update the code in many locations, If you extract it as a shared union type then we can not only reuse that type but also update it in one place only.

already using those in at least two other places in this PR, maybe there's even more so that would be awesome to refactor those as well.

piotrwitek (changes requested):

features/ajna/common/types/AjnaPanel.types.ts
```diff
@@ -1,3 +1,3 @@
-export type AjnaBorrowPanel = 'collateral' | 'quote' | 'switch' | 'close' | 'adjust'
-export type AjnaEarnPanel = 'adjust' | 'liquidity' | 'claim-collateral'
-export type AjnaMultiplyPanel = 'adjust' | 'collateral' | 'quote' | 'switch' | 'close'
+export type BorrowPanel = 'collateral' | 'quote' | 'switch' | 'close' | 'adjust'
+export type earnPanel = 'adjust' | 'liquidity' | 'claim-collateral'
+export type multiplyPanel = 'adjust' | 'collateral' | 'quote' | 'switch' | 'close'
```
piotrwitek, line 3: typos, 1st capital letter

paszkowskiDamian:
As we are trying to come up with a standard for handling lending protocols. What do you think about this crazy idea to forget about `borrow | multiply` positions. How about having a position (probably it would be just current multiplyPosition), as the multiply is just `Summer.fi` name and does not have any influence how the position looks like from the blockchain perspective. The difference would be that the details section and some labels in form. Especially when product wants to add all multiply features to borrow it makes it easer to maintain just one state. @piekczyk @piotrwitek @piotrkonowrocki ?

paszkowskiDamian:
@piekczyk said it's old POC

PR #3001, opened 2023-10-20
Created two different dockerfiles

marcinciarka:
Created two different dockerfiles - prod/staging - so we have more time to test staging build changes (and working prod deployment)


andreidavid (approved):
Use `git mv oldfilename newfilename` to preserve git history. For example Dockerfile to Dockerfile.production.

PR #3002, opened 2023-10-20
✨ New earn strategies

jakubswierczek:
## Changes 👷‍♀️
- New earn strategies: sDAI/USDC and sDAI/USDT with e-mode - default slippage 0.1%. 
- Right now, sDAI/USDT is disabled due to some configuration flaws. It will be fixed in the next PR. 
- Removed unused `protocolData` from `aaveLike` protocols. 
- `OnChainPosition` directly from `actions`
- Added logic to handle max borrow and supply caps in aave-v3 and spark. 
  - Error when deposited amount is bigger than possible supply cup,
  - Error when borrowed amount is bigger than possible borrow cup,
  - Error when the debt position and collateral exceeded the maximum caps with the target LTV, 
- Special components to display those errors. 
- Some cleanup work with local docker setup. 
  
## Known tech debt/errors that will be addressed later.
- The app is not responding when the product hub request fails. The request shouldn't be blocking.
- Prisma - Reset the connection when the database is unavailable. (Try stopping the Docker database and restarting it. You will need to relaunch your app.)
- Display errors when FlashLoan liquidity is exceeded. There are some issues with obtaining that information in Spark.

## How to test 🧪
- Update product hub,
- Try to open a new spark earn strategy,
- Use some big amount to check the errors.
- Try to open a positon.


blockchain/spark-v3/spark-v3-pool-data-provider.ts
```diff
@@ -168,3 +169,27 @@ export function getSparkV3EModeCategoryForAsset({
...
+export interface SparkReserveCapParameters extends BaseParameters {
+  token: string
+}
+
+export interface SparkReserveCap {
+  supply: BigNumber
+  borrow: BigNumber
+}
+
+export function getReserveCaps({
```
piotrwitek, line 182: could we generalize as aavelike?

piotrwitek (changes requested):
Hey. I found a few instances that could be generalized 💪 and left some comment

blockchain/spark-v3/spark-v3-pool-data-provider.ts
jakubswierczek: there is a blockchain call so I renamed it to `getSparkReserveCaps`

piotrwitek (dismissed):

piotrwitek (approved):

PR #3056, opened 2023-10-31
🍿 [Portfolio] First batch of components - header + overview

marcinciarka:
- PortfolioHeader
- PortfolioOverview
- PortfolioLayout
- Simple RWD view for the above

### DIFF
--- a/components/layouts/PortfolioLayout.tsx
+++ b/components/layouts/PortfolioLayout.tsx
@@ -1,0 +1,26 @@
+import { Footer } from 'components/Footer'
+import { ModalTrezorMetamaskEIP1559 } from 'components/Modal'
+import { NavigationController } from 'features/navigation/controls/NavigationController'
+import type { WithChildren } from 'helpers/types/With.types'
+import React from 'react'
+import { BackgroundLight } from 'theme/BackgroundLight'
+
+import { WithAnnouncementLayout } from './WithAnnouncementLayout'
+
+export function PortfolioLayout({ children }: WithChildren) {
+  return (
+    <>
+      <WithAnnouncementLayout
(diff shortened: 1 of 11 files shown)


piotrwitek (approved):
clean code nice

components/layouts/PortfolioLayout.tsx
```diff
@@ -0,0 +1,26 @@
+import { Footer } from 'components/Footer'
+import { ModalTrezorMetamaskEIP1559 } from 'components/Modal'
+import { NavigationController } from 'features/navigation/controls/NavigationController'
+import type { WithChildren } from 'helpers/types/With.types'
+import React from 'react'
+import { BackgroundLight } from 'theme/BackgroundLight'
+
+import { WithAnnouncementLayout } from './WithAnnouncementLayout'
```
piotrkonowrocki, line 8: Please don't use relative imports.

components/layouts/PortfolioLayout.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export function PortfolioLayout({ children }: WithChildren) {
```
piotrkonowrocki, line 10: `WithChildren` is bad pattern introduced a long time ago and repeated so many times in the codebase, but instead of that we should use one of two approaches:
1. `FC` style component that already contains `children` as optional prop,
2. `PropsWithChildren` which does same thing as `WithChildren`, but is native export from React.

components/layouts/PortfolioLayout.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <>
```
piotrkonowrocki, line 12: Unnecessary fragment that wraps single component.

PR #3064, opened 2023-11-02
🍿 [Portfolio] Position components

marcinciarka:
🍿 [Portfolio] Position components


components/portfolio/positions/PortfolioPositionAutomationIcons.tsx
piotrkonowrocki, line 1: I think I saw some designs with automation icons in Product Hub in the table, maybe this component could be places inside generic components without direct connection to Portfolio in its name.

components/GenericSelect.tsx
```diff
@@ -40,6 +40,7 @@ export interface GenericSelectProps {
   options: GenericSelectOption[]
   placeholder?: string
   wrapperSx?: ThemeUIStyleObject
+  iconPosition?: 'left' | 'right'
```
piotrkonowrocki, line 43: Imo we should use select as is, as we have it implemented right now without creating another variant of it for the sake of consistency.

piotrkonowrocki (changes requested):
Please also update imports to absolute path in all of the files.

components/GenericSelect.tsx
marcinciarka: Removed

PR #3087, opened 2023-11-06
🍿 [Portfolio] Fixes

marcinciarka:
# 🍿 [Portfolio] Fixes

- updated blog api version
- changed the data on non-owner view notice banner
- merged every portfolio data source into one hook
- **[BREAKING/CRITICAL]** made the blog dot round again if theres more than two lines of text


piotrwitek (changes requested):
I checked, now its loading 3 calls on refresh, previously only 2.

marcinciarka:
> I checked, now its loading 3 calls on refresh, previously only 2.

Makes sense - cause now it loads positions/overview/assets and doesn't load on tab changes.
It will call 4 times when on non-owner view.

piotrwitek (approved):

PR #3088, opened 2023-11-06
🍿[Portfolio] Lending range component

marcinciarka:
# 🍿[Portfolio] Lending range component

https://github.com/OasisDEX/oasis-borrow/assets/5095527/acc2428c-0d0b-4dcc-814d-8020ea3700d0

### DIFF
--- a/lambdas/src/shared/domain-types.ts
+++ b/lambdas/src/shared/domain-types.ts
@@ -1,3 +1,5 @@
+import { LendingRangeType } from './lending-range'
+
 export enum NetworkNames {
   ethereumMainnet = 'ethereum',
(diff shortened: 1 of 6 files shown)


lambdas/src/shared/domain-types.ts
```diff
@@ -57,7 +59,7 @@ export type DetailsType =
 
 export type PositionDetail = {
   type: DetailsType
-  value: string
+  value: string | LendingRangeType
```
piotrwitek, line 62: do we know all values to get rid of unsafe string?

piotrwitek (approved):

PR #3095, opened 2023-11-07
🍿 [Portfolio] Remove 24 change from the overview

marcinciarka:
# 🍿 [Portfolio] Remove 24 change from the overview


piotrkonowrocki (changes requested):
`subValue` params should be removed from loading state component as well, right now skeleton would show bottom value which does not exist on loaded version.

piotrkonowrocki (approved):

PR #3104, opened 2023-11-09
Added price tickers to positions handlers

piotrkonowrocki:
# Added price tickers to positions handlers
  
## Changes 👷‍♀️

- Added price tickers to be available in each positions handler for Portfolio API.

### DIFF
--- a/handlers/portfolio/positions/index.ts
+++ b/handlers/portfolio/positions/index.ts
@@ -10,10 +10,13 @@ export const portfolioPositionsHandler = async (
   req: NextApiRequest,
 ): Promise<{ positions: PortfolioPosition[]; address: string; error?: string }> => {
   const tickersResponse = await tokenTickers()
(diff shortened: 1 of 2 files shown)


marcinciarka (approved):

handlers/portfolio/positions/index.ts
```diff
@@ -1,15 +1,27 @@
...
   req: NextApiRequest,
 ): Promise<{ positions: PortfolioPosition[]; address: string; error?: string }> => {
+  const tickersResponse = await tokenTickers()
+  const tickers = Object.entries(tickersResponse).reduce<Tickers>(
+    (acc, [key, value]) => ({
+      ...acc,
+      [key.toLowerCase()]: new BigNumber(value),
+    }),
+    {},
+  )
```
piotrwitek, line 19: This should be abstracted and added some comment, with explanation

piotrwitek (changes requested):

PR #3133, opened 2023-11-14
[BE] Method to get positions ready to migrate

piotrwitek:


lambdas/build.mjs
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env zx
+
+import 'zx/globals'
+import { exit } from 'process'
+
+// Constants
+
+const LAMBDA_NAMES = ['portfolio-assets', 'portfolio-migrations', 'portfolio-overview']
```
paszkowskiDamian, line 8: Wouldn't be better if the names of the lambdas be some how inferred from the source?

piotrwitek: @paszkowskiDamian we could change the structure so that lambdas are grouped in a separate folder so we can parse it then sure.

lambdas/lib/portfolio-migrations/src/client.ts
```diff
@@ -0,0 +1,157 @@
...
+import { aavePoolContract } from './abi/aavePoolContract'
+import { decodeBitmapToAssetsAddresses } from './decodeBitmapToAssetsAddresses'
+import { aavePoolDataProviderContract } from './abi/aavePoolDataProviderContract'
+import { aaveOracleContract } from './abi/aaveOracleContract'
+import { USD_DECIMALS } from 'shared/constants'
+import { ProtocolMigrationAssets } from './types'
+import { Address, Network, PortfolioMigrationAsset } from 'shared/domain-types'
+import { createtokenService } from './tokenService'
+
+const supportedChainsIds = ['sepolia']
```
jakubswierczek, line 13: `mainet` `optimism` `aritrum` `base`

lambdas/lib/portfolio-migrations/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  // validate the query
+  let address: Address | undefined
+  try {
+    address = getAddressFromRequest(event)
+  } catch (error) {
+    const message = getDefaultErrorMessage(error)
+    return ResponseBadRequest(message)
+  }
+
+  const rpcUrl =
```
jakubswierczek, line 23: it should come from env I guess

lambdas/lib/portfolio-migrations/src/client.ts
```diff
@@ -0,0 +1,157 @@
...
+    publicClient,
+  })
+  const aaveOracle = getContract({
+    address: aaveOracleContract.address,
+    abi: aaveOracleContract.abi,
+    publicClient,
+  })
+
+  // instantiate address service
+  const tokenService = createtokenService(Network.MAINNET)
```
paszkowskiDamian, line 78: Can we infer the network from the client, I'm not sure what is the exact purpose of `getAssets` and for which network is may be used.  

Ok the more I dig in I don't understand few things. You are iterating over a list of supported networks and invoking `getAssets`, How come `aavePoolContract` gets properly resolved and token service always is using mainnet network so I just wonder how it knows how to resolve proper addresses.

lambdas/lib/portfolio-migrations/src/index.ts
piotrwitek: yes good catch

lambdas/lib/portfolio-migrations/src/client.ts
piotrwitek: good catch

lambdas/lib/shared/src/responses.ts
```diff
@@ -1,14 +1,14 @@
 import type { APIGatewayProxyResultV2 } from 'aws-lambda'
-
 import type { DefaultErrorResponse } from './helper-types'
+import { serialize } from './serialize'
 
 export function createOkBody(obj: Record<string, any>): string {
```
jakubswierczek, line 5: `unknown` will be better than `any`

lambdas/lib/portfolio-migrations/src/client.ts
piotrwitek: this is a bug should be used iteration value from supportedNetworks, I missed that one

jakubswierczek (approved):

lambdas/lib/shared/src/responses.ts
piotrwitek: great sugg

lambdas/lib/portfolio-migrations/src/client.ts
```diff
@@ -0,0 +1,157 @@
...
+  price,
+  priceDecimals,
+}: {
+  balance: bigint
+  balanceDecimals: bigint
+  price: bigint
+  priceDecimals: bigint
+}): number {
+  const value = (balance * price) / 10n ** balanceDecimals
+  return Number(value) / Number(10n ** priceDecimals)
```
paszkowskiDamian, line 156: I'm afraid that we might lose some precision here but If it just for fronted to display that we might be fine. What do you think shout applying the precision at the end 
```
 const value = (balance * price) 
  return Number(value) / Number(10n ** (priceDecimals + balanceDecimals))
```
Or we will be at risk of overflowing `Number(value)`

lambdas/lib/portfolio-migrations/src/client.ts
paszkowskiDamian: And how about the aavePoolContract, it's imported so I don't understand how it knows which address to use.

helpers/clients/portfolio-client.ts
```diff
@@ -56,6 +59,33 @@ export const usePortfolioClient = (baseUrl?: string, headers?: HeadersInit) => {
...
+                symbol: 'USDC',
+                balance: 55008759919n,
+                balanceDecimals: 6n,
+                price: 100000000n,
+                priceDecimals: 8n,
+                usdValue: 55008.759919,
+              },
+            },
+          ],
+        })
```
marcinciarka, line 87: Why isn't this using `PortfolioPosition` type? Theres everything we need (on the frontend) there.

marcinciarka (changes requested):

helpers/clients/portfolio-client.ts
piotrwitek: it's different structure, it's not a position

lambdas/lib/portfolio-migrations/src/client.ts
piotrwitek: @paszkowskiDamian this will overflow that's why I did it in separate steps that should not overflow because price is in 8 decimals always from aave oracle

lambdas/lib/portfolio-migrations/src/client.ts
piotrwitek: > And how about the aavePoolContract, it's imported so I don't understand how it knows which address to use.

this is wrong, I forgot to replace it with addressService. Will fix soon

lambdas/lib/portfolio-migrations/src/client.ts
piotrwitek: done

lambdas/build.mjs
piotrwitek: I'll do it when I move lambdas to monorepo, or in a separate PR here

lambdas/lib/portfolio-migrations/src/client.ts
piotrwitek: Fixed, please review @paszkowskiDamian

paszkowskiDamian: I can still see tokenService always getting mainnet address.

lambdas/lib/portfolio-migrations/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    const supportedChainsIds = event.queryStringParameters?.chainIds
+      ?.split(',')
+      .map(Number) as any as ChainId[] // ?? [
+    //   ChainId.MAINNET,
+    //   ChainId.ARBITRUM,
+    //   ChainId.OPTIMISM,
+    //   ChainId.BASE,
+    //   ChainId.SEPOLIA,
+    // ]
```
paszkowskiDamian, line 45: Shouldn't we actually validate if provided id is supported?

helpers/clients/portfolio-client.ts
```diff
@@ -48,6 +51,33 @@ export const usePortfolioClient = (baseUrl?: string, headers?: HeadersInit) => {
         fetchPortfolioGeneric<PortfolioAssetsResponse>('assets', address),
       fetchPortfolioPositions: (address: string) =>
         fetchPortfolioGeneric<PortfolioPositionsReply>('positions', address),
+      fetchPortfolioMigrations: (address: string): Promise<PortfolioMigrationsResponse> => {
```
jakubswierczek, line 54: I think the client position for migrations should be placed in a separate file since we need to use it on various pages.

lambdas/lib/portfolio-migrations/src/index.ts
piotrwitek: Yes I will fix soon

piotrwitek: done

helpers/clients/portfolio-client.ts
piotrwitek: For now migration epic is on hold but I want to merge the lambdas infra ground work that we did on lambdas as it will be used in other epic that Jacob is working on, is it ok to merge it with expectation that this lambda will not be used?
I have removed client integration anyway so it cant be used by mistake.

PR #3172, opened 2023-11-22
Remove `features/ajna` references

piotrkonowrocki:
# [Remove `features/ajna` references](https://app.shortcut.com/oazo-apps/story/12963/remove-all-feature-ajna-references-in-omnikit)

## Changes 👷‍♀️

- Removed all references and imports of files from old Ajna feature directory (except content cards - this is something for separated PR) and copied files to Omnikit. Right now we should have two completely separated, working implementations and removing old version shouldn't affect Omnikit at all.
- Added index files for all (at least I hope) Omnikit directories.


features/omni-kit/components/OmniValidationWithLink.tsx
```diff
@@ -0,0 +1,66 @@
...
+          values={{ ...item.data, collateralToken, quoteToken, token }}
+        />
+      ),
+    },
+  }))
+
+export const OmniValidationWithLink: FC<OmniValidationWithLinkProps> = ({ name, values }) => {
+  const translationKey = `ajna.validations.${name}`
+
+  const linkMap: { [key: string]: string } = {
```
piekczyk, line 42: these are Ajna specific but yet component is Omni, maybe we should extract this map to const and pass resolved value as prop to this component

piekczyk (changes requested):

piekczyk (approved):

PR #3175, opened 2023-11-23
feature: [sc-13019] Add WBTC-DAI Long strategy to Spark

marcinciarka:
# Add WBTC-DAI Long strategy to Spark

Story details: https://app.shortcut.com/oazo-apps/story/13019/spark-v3-add-wbtc-dai-multiply-strat

### DIFF
--- a/features/aave/strategies/ethereum-spark-v3-strategies.ts
+++ b/features/aave/strategies/ethereum-spark-v3-strategies.ts
@@ -176,7 +176,21 @@ const availableTokenPairs: TokenPairConfig[] = [
     productTypes: {
       [ProductType.Multiply]: {
         featureToggle: undefined,
-        additionalManageActions: [],
+        additionalManageActions: [
+          {
+            action: 'switch-to-borrow',
+            featureToggle: undefined,
+          },
+        ],
+      },
+      [ProductType.Borrow]: {
+        featureToggle: undefined,
+        additionalManageActions: [
+          {
+            action: 'switch-to-multiply',
+            featureToggle: undefined,
+          },
+        ],
       },
     },
   },
--- a/features/aave/strategies/ethereum-spark-v3-strategies.ts
+++ b/features/aave/strategies/ethereum-spark-v3-strategies.ts
@@ -169,6 +169,17 @@ const availableTokenPairs: TokenPairConfig[] = [
       },
     },
   },
+  {
+    collateral: 'WBTC',
+    debt: 'DAI',
+    strategyType: StrategyType.Long,
+    productTypes: {
+      [ProductType.Multiply]: {
+        featureToggle: undefined,
+        additionalManageActions: [],
+      },
+    },
+  },
 ]
 
 const borrowStrategies: IStrategyConfig[] = availableTokenPairs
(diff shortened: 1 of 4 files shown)


piotrkonowrocki (dismissed):

features/aave/strategies/ethereum-spark-v3-strategies.ts
FBrinkkemper, line 1: @marcinciarka what is the additionalManageActions, and would we need it? Ideally we want this strategy for both Multiply & Borrow (I might have not been super clear about that.)

handlers/product-hub/update-handlers/sparkV3/sparkV3Handler.ts
```diff
@@ -158,8 +158,12 @@ export default async function (tickers: Tickers): ProductHubHandlerResponse {
           tokensReserveConfigurationData.find((data) => data[primaryToken]),
         )[primaryToken]
         const tokensAddresses = getNetworkContracts(NetworkIds.MAINNET).tokens
-        // rewards are available for the ETH-like/DAI pairs
-        const hasRewards = primaryTokenGroup === 'ETH' && secondaryToken === 'DAI'
```
FBrinkkemper, line 162: The rewards are due to it having the secondarytoken being DAI. (but there's an exception that you cannot use sDAI as primary token for the rewards).
So I think its:

ETH primary, OR DAI secondary AND NOT sDAI primary.

handlers/product-hub/update-handlers/sparkV3/sparkV3Products.ts
```diff
@@ -145,4 +145,16 @@ export const sparkV3ProductHubProducts: ProductHubItemWithoutAddress[] = [
...
+  {
+    product: [ProductHubProductType.Multiply],
+    primaryToken: 'WBTC',
+    primaryTokenGroup: 'BTC',
+    secondaryToken: 'DAI',
+    depositToken: 'DAI',
+    network: NetworkNames.ethereumMainnet,
+    protocol: LendingProtocol.SparkV3,
+    label: 'WBTC/DAI',
+    multiplyStrategyType: 'short',
```
FBrinkkemper, line 157: why is this short?

FBrinkkemper (changes requested):
Thanks Marcin - posted a few comments :)

features/aave/strategies/ethereum-spark-v3-strategies.ts
marcinciarka: Adding borrow as well. This will result with `additionalManageActions` being `switch-to-borrow` and `switch-to-multiply` then.

handlers/product-hub/update-handlers/sparkV3/sparkV3Handler.ts
marcinciarka: This is checking if the primary token **group** is ETH or BTC (so WSTETH, WBTC works as well) and the literal value of secondary token is DAI (so no SDAI). If we checked `secondaryTokenGroup` that would be more broad).

FBrinkkemper: It is indeed more broad - but your way works right now with the current products too.

FBrinkkemper: ah wait - the group bit. Let's keep it as your way otherwise we might break it :P

FBrinkkemper (approved):

PR #3179, opened 2023-11-24
Fix earnings calculation bug in Aave-like position handler

marcinciarka:
# Fix earnings calculation bug in Aave-like position handler


chrisbradbury88 (approved):
LGTM

PR #3486, opened 2024-02-02
Fixed portfolio styles issue

piotrwitek:
Fixed portfolio styles issue

### DIFF
--- a/components/portfolio/positions/PortfolioPositionBlock.tsx
+++ b/components/portfolio/positions/PortfolioPositionBlock.tsx
@@ -34,7 +34,10 @@ const getColorsPerProtocol = (
         gradientBorder: '#F58013',
       }
     default:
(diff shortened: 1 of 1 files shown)


components/portfolio/positions/PortfolioPositionBlock.tsx
```diff
@@ -34,7 +34,10 @@ const getColorsPerProtocol = (
         gradientBorder: '#F58013',
       }
     default:
-      throw new Error(`Not implemented protocol ${protocol}`)
+      return {
```
piekczyk, line 37: maybe instead of throwing error we could return these defaults + console.warn?

piotrwitek: done

piekczyk (approved):

PR #3636, opened 2024-03-08
ERC-4626 manage overview section view

piotrkonowrocki:
# [ERC-4626 manage overview section view](https://shortcutLink)

image
  
## Changes 👷‍♀️

- Created overview section for ERC-4626 Omnikit protocol implementation - so far full of mocked data.
- Created `Erc4626VaultAllocation` and `Erc4626DetailsSectionContentEstimatedEarnings` components to display Steakhouse specific data.
- Added `headline` prop to Omnikit that allows overwriting default position title.
- Created ERC-4626 dedicated hooks for displaying data in overview section (generic link and vault fee).
- Modified `DetailsSection` component in a way that it can be a accordion now and removed `buttons` props from it. That prop was designed in the beginning as an experiment, but not once used since.
- Modified `ContentCard` component to allow setting up entire card as a link.
- Refactored `DetailsSectionContentTable` component to use `table` instead of `div` displayed as grid. Latest design updates required some changes that were not possible without proper html table.
- Refined the way tooltips in `VaultHeadlineDetails` component are displayed, according to latest design.
- Added compact (paddless) mode to `AssetsTable` component.
- Added `MORPHO` token to the config (no price feed available).
- Moved some Ajna specific translations under `omni-kit` key as now they are reused across different protocol implementations.
  
## How to test 🧪

 - `/ethereum/erc-4626/earn/steakhouse-USDC` address should load fine with partial UI and mocked data.


components/DetailsSection.tsx
```diff
@@ -22,8 +21,9 @@ interface DetailsSectionProps {
 }
 
 export function DetailsSection({
+  accordion,
```
jakubswierczek, line 24: ? :D

jakubswierczek (approved):

PR #3651, opened 2024-03-12
chore: merge duplicated events in favor of automation

halaprix:

image
  
## Changes 👷‍♀️
- if there are two history items with the same hash - use only the first one - or an `Autoamtion` related event if available
## How to test 🧪
- http://localhost:3000/base/aave/v3/304#history


features/positionHistory/filterAndGroupByTxHash.ts
```diff
@@ -0,0 +1,17 @@
+import { find, groupBy, map } from 'lodash'
+
+/**
+ * Filters and groups events by transaction hash.
+ *
+ * @param events - The array of events to filter and group.
+ * @returns An array of events, where each event is the first event in a group with the same transaction hash,
+ *          or the first event with 'Automation' in its kind if such event exists in the group.
+ */
+export function filterAndGroupByTxHash(events: any[]): any[] {
```
jakubswierczek, line 10: Why are events typed as `any`? If we don't know the type, we should use `unknown`.

jakubswierczek: Make it generic maybe `T extend { txHash: string, kind: string }` or something like that.

jakubswierczek (changes requested):

jakubswierczek (approved):

PR #3680, opened 2024-03-18
feat: [sc-14820] [FE] Integrate Claims & Rewards functions

jakubswierczek:
Story details: https://app.shortcut.com/oazo-apps/story/14820

The getErc4626ApyParameters function is refactored to incorporate new interfaces and fetch real-time data over hardcoded values. This refactoring enriches the response data handling various reward tokens and market situations. Alongside, a rewrite rule is added to the server configuration, redirecting '/api/morpho' requests to the correct API URL.

Updated the function fetching token prices in prices.constants.ts to use the Fetch API instead of ajax from rxjs. This also involves updating the "@oasisdex/dma-library" to version 0.6.35 and adjusting the getErc4626ApyParameters function to work with the real token prices data. Additionally, a new getErc4626Claims helper function was introduced for fetching claim details.

### DIFF
--- a/features/omni-kit/protocols/erc-4626/helpers/getErc4626ApyParameters.ts
+++ b/features/omni-kit/protocols/erc-4626/helpers/getErc4626ApyParameters.ts
@@ -62,31 +62,40 @@ export async function getErc4626ApyParameters(
     return {
       vault: {
         apy: '0',
-        curator: 'Steakhouse',
         fee: '0',
       },
       apyFromRewards: [],
       allocations: [],
     }
   }
+
   const data = (await response.json()) as GetRewardsResponse
+
+  const apyFromRewards = data.rewardsByToken
+    .map((token) => {
+      return {
+        token: token.symbol.toUpperCase(),
+        value: token.apy.toString(),
+        per1kUsd: token.humanReadable.toString(),
+      }
+    })
+    .filter((token) => token.value !== '0')
+
+  const allocations = data.rewardsByMarket
+    .map((market) => ({
+      token: market.market.collateral.symbol.toUpperCase(),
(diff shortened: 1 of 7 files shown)


features/omni-kit/protocols/erc-4626/helpers/getErc4626ApyParameters.ts
```diff
@@ -1,30 +1,92 @@
...
+    `/api/morpho/meta-morpho?address=${vaultAddress}&morhoPrice=1&wsEthPrice=${wstETHPrice}`,
+  )
+
+  if (response.status !== 200) {
+    console.warn('Failed to fetch vault details', response)
+    return {
       vault: {
-        apy: '0.0175',
+        apy: '0',
         curator: 'Steakhouse',
```
piotrkonowrocki, line 65: `curator` is optional, so in case there's an error we can just skip it instead of defaulting to Steakhouse.

features/omni-kit/protocols/erc-4626/helpers/getErc4626ApyParameters.ts
```diff
@@ -1,30 +1,92 @@
...
-          value: '0.025',
-          per1kUsd: '250',
-        },
-      ],
+      apyFromRewards: [],
+      allocations: [],
+    }
+  }
+  const data = (await response.json()) as GetRewardsResponse
+  return {
```
piotrkonowrocki, line 73: Could you please change the return in a way that if `rewardsByToken.lenght === 0`, then `apyFromRewards` and `allocations` aren't returner at all instead of being returned with `[]` as value?

piotrkonowrocki (changes requested):

piotrkonowrocki (approved):

PR #3688, opened 2024-03-19
feat: [sc-14862] [FE] History for ERC-4626

jakubswierczek:
Story details: https://app.shortcut.com/oazo-apps/story/14862

This commit adds handling and mapping of ERC4626 history events. It includes adding methods for fetching aggregated data, mapping between different events types and updates in displaying event labels. With this update, the application now supports more complete and accurate tracking of transaction history for ERC4626 protocol.

CleanShot 2024-03-19 at 12 40 20@2x

### DIFF
--- a/features/subgraphLoader/consts.ts
+++ b/features/subgraphLoader/consts.ts
@@ -675,6 +675,75 @@ export const subgraphMethodsRecord: SubgraphMethodsRecord = {
       }
     }
   `,
+  getErc4626PositionAggregatedData: gql`
+    query Erc4626AggregatedData($vault: String!, $dpmProxyAddress: String!) {
+      summerEvents(where: { account: $dpmProxyAddress, vault: $vault }) {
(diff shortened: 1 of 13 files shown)


features/subgraphLoader/consts.ts
```diff
@@ -675,6 +675,75 @@ export const subgraphMethodsRecord: SubgraphMethodsRecord = {
       }
     }
   `,
+  getErc4626PositionAggregatedData: gql`
+    query Erc4626AggregatedData($vault: String!, $dpmProxyAddress: String!) {
+      summerEvents(where: { account: $dpmProxyAddress, vault: $vault }) {
```
piekczyk, line 680: will these filter params be enough to get events per given position without mixing up events from few different positions (assuming that one dpm can have multiple positions)

piekczyk (approved):

PR #3781, opened 2024-04-11
feat: [sc-15119] [CI] Enable ENA Claiming from SUSDE/USDE positions where ENA is present on Proxy

jakubswierczek:
Story details: https://app.shortcut.com/oazo-apps/story/15119

A new feature has been added to support claimable ERC20 tokens. This includes the ability to fetch token balances and encode transfer proxy actions. Changes also include refactoring component names and correcting translations. A new section displaying vault token rewards has been introduced and a bug fixed where missing items in rows result in errors.


blockchain/better-calls/erc20.ts
```diff
@@ -145,3 +156,47 @@ export async function createDisapproveTransaction({
   const contract = Erc20__factory.connect(tokens[token].address, signer)
   return await contract.approve(spender, 0)
 }
+
+export async function encodeTransferProxyAction({
```
halaprix, line 160: this could be rfactored to accept `to` address as parameter - then another method would use this method passing dpm owner as transfer target address

OR

rename this to `encodeTransferToOwnerProxyAction`

piekczyk (approved):

halaprix (approved):

PR #3839, opened 2024-04-26
Update Ajna Rewards Pools 25-4-24

chrisbradbury88:
Updating the Ajna rewards and pools

### DIFF
--- a/features/omni-kit/protocols/ajna/helpers/isPoolWithRewards.ts
+++ b/features/omni-kit/protocols/ajna/helpers/isPoolWithRewards.ts
@@ -101,16 +101,6 @@ export const ajnaWeeklyRewards: AjnaWeeklyRewards = {
       borrowShare: new BigNumber(0.05),
       earnShare: new BigNumber(0.95),
     },
-    'ENA-SDAI': {
-      amount: new BigNumber(0),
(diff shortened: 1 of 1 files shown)


features/omni-kit/protocols/ajna/helpers/isPoolWithRewards.ts
```diff
@@ -92,34 +92,34 @@ export const ajnaWeeklyRewards: AjnaWeeklyRewards = {
...
     },
     'MWSTETHWPUNKS40-WSTETH': {
-      amount: new BigNumber(13120),
+      amount: new BigNumber(6560),
       borrowShare: new BigNumber(0.05),
       earnShare: new BigNumber(0.95),
     },
     'ENA-SDAI': {
-      amount: new BigNumber(4920),
+      amount: new BigNumber(0),
```
piotrkonowrocki, line 105: If there are no rewards for a pool, entire block should be removed from config. If we leave it in the config it's still going to be shown on Product Hub with sparks icon, but with 0 weekly and 0% in tooltip.

piotrkonowrocki (changes requested):

piotrkonowrocki (approved):

PR #3906, opened 2024-05-24
Added SDK_API_URL

piotrwitek:


server/rewrites.ts
```diff
@@ -18,10 +18,6 @@ const rewriteRules = () => [
     source: '/api/migrations',
     destination: `${process.env.FUNCTIONS_API_URL}/api/migrations`,
   },
-  {
```
robercano, line 21: Don't we need a rewrite for `api/sdk`?

robercano (approved):
Looks good! Just a question about whether we still need the rewrite

server/rewrites.ts
piotrwitek: no, it should go through proxy endpoint, rewrite is bad it'll put the URL in the header making it fully visible to everybody (expose our endpoint) and also fails requests because of breaking header size limits due to long URI

PR #3918, opened 2024-06-03
Aave like dpm position history split

piekczyk:
# Aave like dpm position history split

## Changes 👷‍♀️

- handled recent change in subgraph where we are able to distinguish between aavelike positions on the same dpm
  
## How to test 🧪

- spark & aave history should be separated in subgraph response even if on the same dpm

### DIFF
--- a/features/omni-kit/protocols/spark/settings.ts
+++ b/features/omni-kit/protocols/spark/settings.ts
@@ -12,9 +12,11 @@ const availableSparkAutomations = [
   AutomationFeatures.PARTIAL_TAKE_PROFIT,
 ]
 
+export const sparkRawProtocolName = 'Spark'
+
 export const settings: OmniProtocolSettings = {
   rawName: {
-    [NetworkIds.MAINNET]: 'Spark',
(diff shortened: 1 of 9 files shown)


features/omni-kit/protocols/spark/settings.ts
```diff
@@ -12,9 +12,11 @@ const availableSparkAutomations = [
   AutomationFeatures.PARTIAL_TAKE_PROFIT,
 ]
 
+export const sparkRawProtocolName = 'Spark'
```
halaprix, line 15: It may be more efficient to relocate the raw protocol names to an enum situated with the other network name definitions in the same file. This way, they can be imported and utilized here as needed.

It's up to you ;)

halaprix (approved):

features/omni-kit/protocols/spark/settings.ts
piekczyk: I would rather keep it as const for now since `raw names` may be different per specific protocol (ajna) and vary between for networks

PR #3920, opened 2024-06-03
Aave like dpm reusability adjustment

piekczyk:
# Aave like dpm reusability adjustment

## Changes 👷‍♀️

- allow to reuse dpm proxy if there is only spark or aave position on given dpm and new position will be of different protocol
  
## How to test 🧪

- if there is already created spark position on dpm, it should be possible to use this dpm to open aave position or other way around


piotrwitek (approved):
looks good

PR #3926, opened 2024-06-04
Emode upper case patch

piotrwitek:
To fix Still same Asset not found error for all Aave V3 token when refinancing


robercano (approved):
Looks good!

PR #3933, opened 2024-06-06
Update 1inch API header

piotrwitek:
Updated 1inch API header


pages/api/exchange/[[...params]].tsx
```diff
@@ -4,7 +4,7 @@ import { config } from 'server/config'
 import { z } from 'zod'
 
 const ROUTE_NAME = 'exchange'
-const AUTH_HEADER_KEY = 'auth-key'
```
Comment (piekczyk) [inline pages/api/exchange/[[...params]].tsx:7]: on line 74 we need to also change 

```const headers = { [AUTH_HEADER_KEY]: 'Bearer ' + oneInchApiKey }```

piekczyk (changes requested):

piekczyk (approved):

PR #3943, opened 2024-06-07
removed LUSD from emode stablecoin tokens list

piotrwitek:
removed LUSD from emode stablecoin tokens list to set correct emode in SDK


robercano (approved):
Looks good!

PR #3971, opened 2024-06-26
"Implement separate netApy and borrow costs comparison for refinance"

piotrwitek:
Previously, the borrow rate was used to calculate the net APY for refinance positions. This pull request introduces a separate netApy field that accurately represents the net APY based on the current LTV. The borrow rate is now used solely for calculating the increase in debt over one year at the current market utilization. This change improves the accuracy of the net APY calculation and provides more meaningful information to users.

### DIFF
--- a/features/productHub/types.ts
+++ b/features/productHub/types.ts
@@ -77,6 +77,7 @@ export interface ProductHubItemDetails {
   earnStrategyDescription?: string
   // borrow rate
   fee?: string
+  netApy?: string
   hasRewards?: boolean
   liquidity?: string
   managementType?: ProductHubManagementType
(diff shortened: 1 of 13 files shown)


robercano (dismissed):
Looks good to me but I'd suggest to wait for Seba review as well if possible

features/productHub/types.ts
```diff
@@ -77,6 +77,7 @@ export interface ProductHubItemDetails {
   earnStrategyDescription?: string
   // borrow rate
   fee?: string
+  netApy?: string
```
piekczyk, line 80: Is this available in db? If not it shouldn't be in this interface

piekczyk (changes requested):

features/productHub/types.ts
piotrwitek: good catch it's fixed

piekczyk (approved):

PR #4042, opened 2024-10-08
Fixed swap on mainnet if WBTC we need to remove CURVE from LP

piotrwitek:

Fix big swaps of WBTC (> 1BTC), to not fail
  
## Changes 👷‍♀️
on mainnet if WBTC we need to remove CURVE from LP
  
## How to test 🧪
create an ETH/WBTC position, deposit at least 250 ETH and try to take debt to swap at least several BTCs

## Summary by CodeRabbit

- **New Features**
	- Enhanced the swapping logic to filter out the 'CURVE' protocol for WBTC on the MAINNET, improving the accuracy of protocol selection during token swaps.

- **Bug Fixes**
	- Refined the control flow for determining protocols, ensuring better handling of specific token and network combinations.

### DIFF
--- a/helpers/swap.ts
+++ b/helpers/swap.ts
@@ -98,7 +98,7 @@ export function getOneInchCall(
     slippage: BigNumber,
     protocols: string[] = [],
   ) => {
-    const resolvedProcotols = match({ protocols, networkId })
+    let resolvedProcotols = match({ protocols, networkId })
       .with(
         { protocols: [], networkId: NetworkIds.OPTIMISMMAINNET },
         () => OPTIMISM_DEFAULT_PROCOTOLS,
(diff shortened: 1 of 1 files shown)


helpers/swap.ts
```diff
@@ -117,6 +117,11 @@ export function getOneInchCall(
       )
       .otherwise(() => protocols)
 
+    // on mainnet if WBTC we need to remove CURVE from LP
+    if (networkId === NetworkIds.MAINNET && from === '0x2260fac5e5542a773aa44fbcfedf7c193bc2c599') {
```
piekczyk, line 121: just to be sure I would add `from.toLowerCase()`

piotrwitek: done

